PR #11, opened 2016-07-21
Large Barrel example code review

alexkernphysiker:


LargeBarrelAnalysis/TaskB1.h
```diff
@@ -17,44 +17,33 @@
...
+#include <JPetTask/JPetTask.h>
+#include <JPetRawSignal/JPetRawSignal.h>
+#include <JPetTimeWindow/JPetTimeWindow.h>
+#include <JPetParamBank/JPetParamBank.h>
+#include <JPetParamManager/JPetParamManager.h>
 #include "LargeBarrelMapping.h"
-
 class JPetWriter;
-
+#ifdef __CINT__
```
alekgajos: I think it would be good to add a comment here explaining why this is necessary. At first glance it looks rather cryptic.

alekgajos:
I have two general remarks:

1. The "#define override" lines deserve a comment explaining why this is necessary as this looks a bit cryptic at first glance.

2. You have removed many of the comments in the example modules. If the comments are wrong, please fix them rather than remove. Some of the comments might have been obvious, but we should keep in mind that this is an example which may be read by beginners who might not even know C++ completely, not to mention the details of the Framework. Even obvious comments are harmless and may help someone to get through this example, so I believe we should tend to add even more comments and not remove the existing ones.

alexkernphysiker:
1) why the instruction #ifdef __CINT__ does not explain it?
2) comments like this one:
-  // get number of TDC channels which fired in one TSlot
-  int ntdc = evt->GetTotalNTDCChannels();
don't make code more clear but create barriers for seeing easily the structure of code
3) there is no document of coding convention in our team to strictly formalize this things

wkrzemien:
You can add the comment before the first usage of define overide a comment like:
//CINT is not able to interpret override keyword therefore we replace is it by the empty string if CINT is 
//used as a compiler
//override meaning can be seen here http://en.cppreference.com/w/cpp/keyword/override

alexkernphysiker:
I don't see how can I update the pull request from here. I've made a push into my fork with adding comments.

PR #15, opened 2016-12-04
Update TaskC.cpp

Alvarness:
Hits are sorted by time (within time window) and then saved to a file.

### DIFF
--- a/LargeBarrelAnalysis/TaskC.cpp
+++ b/LargeBarrelAnalysis/TaskC.cpp
@@ -93,14 +93,17 @@ void TaskC::terminate(){
 }
 
 
-void TaskC::saveHits(const vector<JPetHit>&hits){
-	assert(fWriter);
-	for (auto hit : hits){
-		// here one can impose any conditions on hits that should be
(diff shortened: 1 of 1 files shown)


LargeBarrelAnalysis/TaskC.cpp
```diff
@@ -93,14 +93,17 @@ void TaskC::terminate(){
...
-void TaskC::saveHits(const vector<JPetHit>&hits){
-	assert(fWriter);
-	for (auto hit : hits){
-		// here one can impose any conditions on hits that should be
-		// saved or skipped
-		// for now, all hits are written to the output file
-		// without checking anything
-		fWriter->write(hit);
-	}
+void ModuleC::saveHits(const vector<JPetHit>&hits){
```
wkrzemien: ModuleC -> TaskC.
Since the sorting part of the code is identical with the one used in the JPetEvent, I will extract it to some external function in framework and just call the here.. Maybe we can add some AnalysisTools, or JPetHitUtils, or something like that.

PR #16, opened 2016-12-12
Minor updates of the examples

alekgajos:
Changes:
1. removed reinterpret_cast usages as they are no longer needed
2. removed debugging couts from AnalysisExample
3. CHanged LargeBarrelAnalysis to process all events rather than ends of time windows
4. Added sorting of hits in LargeBarrelAnalysis/TaskC
5. Update version of the framework submodule

### DIFF
--- a/LargeBarrelAnalysis/TaskA.cpp
+++ b/LargeBarrelAnalysis/TaskA.cpp
@@ -50,8 +50,8 @@ void TaskA::exec(){
 			// iterate over all signals from one TDC channel
 			// analyze number of hits per channel
 			getStatistics().getHisto1D("HitsPerEvtCh").Fill( tdcChannel->GetHitsNum() );
-
-			for(int j = 0; j < tdcChannel->GetHitsNum(); ++j){
+			const int kNumHits = tdcChannel->GetHitsNum();
+			for(int j = 0; j < kNumHits; ++j){
 
 			  // check for unreasable times
 			  // the times should be negative (measured w.r.t end of time window)
--- a/LargeBarrelAnalysis/TaskA.cpp
+++ b/LargeBarrelAnalysis/TaskA.cpp
@@ -21,9 +21,10 @@ void TaskA::init(const JPetTaskInterface::Options& opts){
 	getStatistics().createHistogram( new TH1F("HitsPerEvtCh","Hits per channel in one event",50,-0.5,49.5) );
 	getStatistics().createHistogram( new TH1F("ChannelsPerEvt","Channels fired in one event",200,-0.5,199.5) );
 }
+
 TaskA::~TaskA(){}
 void TaskA::exec(){  
-	//getting the data from event in propriate format
+	//getting the data from event in apropriate format
 	//const is commented because this class has inproper architecture:
 	// all get-methods aren't tagged with const modifier
 	if(auto evt = dynamic_cast</*const*/ EventIII*const>(getEvent())){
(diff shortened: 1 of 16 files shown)


LargeBarrelAnalysis/TaskA.cpp
```diff
@@ -49,34 +49,43 @@ void TaskA::exec(){
...
-				// check for empty TDC times
-				if( tdcChannel->GetLeadTime(j) == -100000 )continue;
-				if( tdcChannel->GetTrailTime(j) == -100000 )continue;
-				// finally, set the times in ps [raw times are in ns]
-				sigChTmpLead.setValue(tdcChannel->GetLeadTime(j) * 1000.);
-				sigChTmpTrail.setValue(tdcChannel->GetTrailTime(j) * 1000.);
-				tslot.addCh(sigChTmpLead);
-				tslot.addCh(sigChTmpTrail);
+
+			for(int j = 0; j < tdcChannel->GetHitsNum(); ++j){
```
wkrzemien: I think GetHitsNum should be called once outside the loop otherwise you call this function in every loop iteration.
Maybe it is optimized by the compiler, but I dont know.

LargeBarrelAnalysis/TaskA.cpp
```diff
@@ -49,34 +49,43 @@ void TaskA::exec(){
...
+			  if( tdcChannel->GetLeadTime(j) == -100000 )continue;
+			  if( tdcChannel->GetTrailTime(j) == -100000 )continue;
+
+			  // check for unreasable times
+			  // the times should be negative (measured w.r.t end of time window)
+			  // and not smaller than -1*timeWindowWidth (which can vary for different)
+			  // data but shoudl not exceed 1 ms, i.e. 1.e9 ps)
+			  if( tdcChannel->GetLeadTime(j) > 0 || tdcChannel->GetLeadTime(j) < -1.e9 )continue;
+			  if( tdcChannel->GetTrailTime(j) > 0 || tdcChannel->GetLeadTime(j) < -1.e9  )continue;
+
```
wkrzemien: Is this condition correct? you have trail and lead times check together

LargeBarrelAnalysis/TaskA.cpp
```diff
@@ -49,34 +49,43 @@ void TaskA::exec(){
...
+			  JPetSigCh sigChTmpLead, sigChTmpTrail;
+			  sigChTmpLead.setDAQch(tomb_number);
```
wkrzemien: Maybe, define some helper function to set all those sigCh ?
void setSigChFromTOMB(int tomb_number, const ..& tomb_channel, JPetSigCh::signalEdge edge)
or JPetSigCh generateSigChFromTOMB () 
and here just call
auto sigchTmpLead = generateSigChFromTOMB(tomb_number, tomb_channel, JPetSigCh::Leading);
auto sigchTmpLead = generateSigChFromTOMB(tomb_number, tomb_channel, JPetSigCh::Trailing);

LargeBarrelAnalysis/TaskC.cpp
```diff
@@ -16,10 +16,37 @@
...
+
+
 TaskC::TaskC(const char * name, const char * description):JPetTask(name, description){}
 TaskC::~TaskC(){}
-void TaskC::init(const JPetTaskInterface::Options& opts){}
+
+void TaskC::init(const JPetTaskInterface::Options& opts){
+
+
+  for(int i=1;i<=4;++i){
```
wkrzemien: 4 means number of thresholds ?
maybe again define some const int kNumOfThresholds ?

LargeBarrelAnalysis/TaskC.cpp
```diff
@@ -69,13 +101,47 @@ vector<JPetHit> TaskC::createHits(const vector<JPetRawSignal>&signals){
 					WARNING("TWO hits on the same scintillator side we ignore it");         
 					continue;
 				}
+				
+				if( recoSignalA.getRawSignal().getNumberOfPoints(JPetSigCh::Leading) < 4 ) continue;
+				if( recoSignalB.getRawSignal().getNumberOfPoints(JPetSigCh::Leading) < 4 ) continue;
+
+				bool thresholds_ok = true;
+				for(int i=1;i<=4;++i){
```
wkrzemien: 4 number of thresholds? to const ?

LargeBarrelAnalysis/TaskC.cpp
```diff
@@ -85,17 +151,60 @@ vector<JPetHit> TaskC::createHits(const vector<JPetRawSignal>&signals){
...
+
+
+  
+
+}
+
 void TaskC::saveHits(const vector<JPetHit>&hits){
 	assert(fWriter);
-	for (auto hit : hits){
+      	for (auto hit : hits){
```
wkrzemien: we should use for (auto const & hit : hits) for readonly 
and for (auto hit: hits)  only if we want to work on copies

LargeBarrelAnalysis/TaskE.cpp
```diff
@@ -48,6 +48,15 @@ void TaskE::init(const JPetTaskInterface::Options& opts){
 			}
 		}
 	}
+
+	// create dt histos for each strip
+	for(auto & scin : getParamBank().getScintillators()){
```
wkrzemien: or auto const & scint 
if readonly

LargeBarrelAnalysis/TaskA.cpp
wkrzemien: const auto kHitsNum =tdcChannel->GetHitsNum();
for(int j = 0; j < kHitsNum  ++j) {

LargeBarrelAnalysis/TaskA.cpp
```diff
@@ -99,3 +92,17 @@ const JPetParamBank& TaskA::getParamBank()const{
...
+  JPetSigCh sigch;
+  sigch.setDAQch(channel.getChannel());
+  sigch.setType(edge);
+  sigch.setThresholdNumber(channel.getLocalChannelNumber());
+  sigch.setThreshold(channel.getThreshold());
+  sigch.setPM(channel.getPM());
+  sigch.setFEB(channel.getFEB());
+  sigch.setTRB(channel.getTRB());
+  sigch.setTOMBChannel(channel);
+
```
wkrzemien: return is missing

alekgajos: Thanks for noticing. It turned out that we didn't have compiler warnings enabled for the examples so that I didn't catch that. I have enabled warnings now.

PR #18, opened 2016-12-21
Hit matching

Alvarness:


LargeBarrel/ModuleD.cpp
```diff
@@ -0,0 +1,112 @@
...
+            if (i -> getRecoSignal().getRawSignal().getPM().getScin() == j -> getRecoSignal().getRawSignal().getPM().getScin()) {
+                // found 2 signals from the same scintillator
+
+                // assign sides A and B properly
+
+                if(
+                    (i->getRecoSignal().getRawSignal().getPM().getSide() == JPetPM::SideA)
+                    &&(j->getRecoSignal().getRawSignal().getPM().getSide() == JPetPM::SideB)
+                ){
+                    if( abs(i -> getTime() - j -> getTime()) < TIME_WINDOW /*ps*/){
```
wkrzemien: We have convention to call constants: kTimeWindow 
You can define it at the class level but anyway set the argument to this function

LargeBarrel/ModuleD.cpp
```diff
@@ -0,0 +1,112 @@
...
+            }
+        }
+    }
+}
+
+vector<JPetHit> ModuleD::createHits(const vector<JPetPhysSignal>&signals){
+
+    vector<JPetHit> hits;
+
+    for (auto i = signals.begin(); i != signals.end(); ++i) {
```
wkrzemien: so here instead of double loop, you would loop over your map 
so efectively over the scintilators which have some signals assigned

LargeBarrel/ModuleD.cpp
```diff
@@ -0,0 +1,112 @@
...
+ */
+
+#include <iostream>
+#include <JPetWriter/JPetWriter.h>
+#include <JPetAnalysisTools/JPetAnalysisTools.h>
+#include "ModuleD.h"
+
+using namespace std;
+
+ModuleD::ModuleD(const char * name, const char * description):JPetTask(name, description){}
```
Alvarness: Descriptions of the modules are inside of main function - names of modules are meant to tell position in analysis sequence (I can also add description inside this module)

LargeBarrel/ModuleD.cpp
```diff
@@ -0,0 +1,112 @@
...
+            else {
+                saveHits(createHits(fSignals));
+                fSignals.clear();
+                fSignals.push_back(*currSignal);
+            }
+        }
+    }
+}
+
+vector<JPetHit> ModuleD::createHits(const vector<JPetPhysSignal>&signals){
```
Alvarness: Done

LargeBarrel/ModuleD.cpp
Alvarness: Done

LargeBarrel/ModuleD.cpp
```diff
@@ -0,0 +1,112 @@
...
+ModuleD::~ModuleD(){}
+
+void ModuleD::init(const JPetTaskInterface::Options& opts){
+}
+
+void ModuleD::exec(){
+    //getting the data from event in propriate format
+    if(auto currSignal = dynamic_cast<const JPetPhysSignal*const>(getEvent())){
+
+        if (fSignals.empty()) {
```
Alvarness: Now we have a map of scintillators, and two vectors connected with each scintillator. Those vectors contains signals from side A and B - then we try to pair signal from side A with a signal from side B - this approach should be faster then previous one

LargeBarrel/ModuleD.cpp
wkrzemien: Ok, but this class will become at the end the module of the j-pet-framework, and someone could use it in his analysis chain without all other modules - so it must have a class name with some descriptive name. ModuleD doesnt explain anything. Descriptive names are important

LargeBarrel/ModuleD.cpp
```diff
@@ -0,0 +1,120 @@
...
+            else {
+                saveHits(createHits( fAllSignalsInTimeWindow, kTimeWindow));
+                fAllSignalsInTimeWindow.clear();
+                fillSignalsMap(*currSignal);
+            }
+        }
+    }
+}
+
+vector<JPetHit> ModuleD::createHits(map<int, pair<vector<JPetPhysSignal>, vector<JPetPhysSignal>>> fAllSignalsInTimeWindow, const double kTimeWindow){
```
wkrzemien: should be createHits( const map ...& signalsInTimeWindow, double timeWindow)
passing by reference should be a default for objects or containers
Arguments of the functions should not have the same names as members, especially you dont want to use fAllSignalsIn... or kDefaultWindow as local agruments.
Since you have two notions of timeWindow (one time window for hits - your kTimeWindow, second from DAQ) you shoud change the name e.g. use hitTimeWindow and e.g. DAQTWindow and add comments, because it will be misleading

LargeBarrel/ModuleD.h
```diff
@@ -0,0 +1,57 @@
...
+        ModuleD(const char * name, const char * description);
+        virtual ~ModuleD();
+        virtual void init(const JPetTaskInterface::Options& opts)override;
+        virtual void exec()override;
+        virtual void terminate()override;
+        virtual void setWriter(JPetWriter* writer)override;
+
+    protected:
+        bool isFirstSignalSet = false;
+        int timeWindowIndex;
```
wkrzemien: int timeWindowIndex = -1;   --> or other value that can be treated as not_set

LargeBarrel/ModuleD.h
```diff
@@ -0,0 +1,57 @@
...
+
+        std::map <int, std::pair <std::vector<JPetPhysSignal>, std::vector<JPetPhysSignal>> > fAllSignalsInTimeWindow;
+        // Map od all signals within a single time window.
```
wkrzemien: This comment should be above the std::map, and also use triple /// 
/// is recognised by Doxygen.

To make the definitions shorter I would use 
typedef for std::map <int, ....>

Make more explicit which time window do you mean (DAQ or yours)

LargeBarrel/ModuleD.h
```diff
@@ -0,0 +1,57 @@
...
+        // It has a structure of < ScintillatorID <SignalsOnSideA, SignalsOnSideB> >
+
+        void fillSignalsMap(JPetPhysSignal signal);
+        std::vector<JPetHit> createHits(std::map <int, std::pair <std::vector<JPetPhysSignal>, std::vector<JPetPhysSignal>> > fAllSignalsInTimeWindow, const double kTimeWindow);
+        void saveHits(const std::vector<JPetHit>&hits);
+
+        JPetWriter* fWriter;
```
wkrzemien: JPetWriter* fWriter = 0;   or  = nullptr;

PR #19, opened 2016-12-26
Hit matching

Alvarness:
Comments added. Modules names changed to more meaningful rather then ModuleX


LargeBarrel/conf_trb3.xml
```diff
@@ -0,0 +1,353 @@
+<?xml version="1.0"?>
```
wkrzemien: what is this?
The config files and data files should be separated from the code. We have a special place to it on sphinx

LargeBarrel/conf_trb3.xml
```diff
@@ -0,0 +1,353 @@
...
+<READOUT>
```
wkrzemien: What is this? 
Config files should be places at sphinx.

LargeBarrel/Translator.h
```diff
@@ -0,0 +1,65 @@
...
+        virtual void exec()override;
+        virtual void terminate()override;
+        virtual void setWriter(JPetWriter* writer)override;
+        virtual void setParamManager(JPetParamManager* paramManager)override;
+        const JPetParamBank& getParamBank()const;
+
+    protected:
+        void saveTimeWindow( JPetTimeWindow slot);
+        JPetSigCh generateSigCh(const JPetTOMBChannel & channel, JPetSigCh::EdgeType edge) const;
+        JPetWriter* fWriter;
```
wkrzemien: fWriter = 0;

LargeBarrel/Translator.h
```diff
@@ -0,0 +1,65 @@
...
+        JPetParamManager* fParamManager;
+        long long int fCurrEventNumber;
+
+        const double kMaxTime = 0.;     //
```
wkrzemien: Sorry still not very obvious - add one line command to explain that it is a max time of .... and min time of ...

LargeBarrel/HitFinder.h
```diff
@@ -0,0 +1,74 @@
...
+         * described in class descrption */
+        typedef std::map <int, std::pair <std::vector<JPetPhysSignal>, std::vector<JPetPhysSignal>> > SignalsContainer;
+
+        int DAQTimeWindowIndex = -1; /// Index which defines a given DAQ time window (defined at the hardware level).
+
+        SignalsContainer fAllSignalsInTimeWindow;
+
+
+        void fillSignalsMap(JPetPhysSignal signal);
+        std::vector<JPetHit> createHits(const SignalsContainer& fAllSignalsInTimeWindow, const double kTimeDifferenceWindow);
```
wkrzemien: the local parameters of the function should not be call the same as class members. 

So:
std::vector<JPetHit> createHits(const SignalsContainer& fAllSignalsInTimeWindow, const double kTimeDifferenceWindow);
to:
std::vector<JPetHit> createHits(const SignalsContainer& allSignalsInTimeWindow, const double timeDifferenceWindow);

LargeBarrel/HitFinder.cpp
```diff
@@ -0,0 +1,123 @@
...
+                saveHits(createHits( fAllSignalsInTimeWindow, kTimeWindowWidth));
+                fAllSignalsInTimeWindow.clear();
+                fillSignalsMap(*currSignal);
+            }
+        }
+    }
+}
+
+
+vector<JPetHit> HitFinder::createHits(const SignalsContainer& fAllSignalsInTimeWindow, const double kTimeDifferenceWindow){
```
wkrzemien: The same as in header case you should call the local arguments of the function differently than fAll... or kTime...
e.g. allSignal and timeDiffe

LargeBarrel/HitFinder.cpp
```diff
@@ -0,0 +1,123 @@
...
+
+void HitFinder::terminate(){
+    saveHits(createHits(fAllSignalsInTimeWindow, kTimeWindowWidth)); //if there is something left
+}
+
+
+void HitFinder::saveHits(const vector<JPetHit>& hits){
+    assert(fWriter);
+    JPetAnalysisTools sort;
+    auto sortedHits = sort.getHitsOrderedByTime(hits);
```
wkrzemien: auto sortedHits = JPetAnalysisTools::getHitsOrderedByTime(hits);
This syntax explicitely shows that getHitsOrdereByTime is a static function.

LargeBarrel/HitFinder.cpp
```diff
@@ -0,0 +1,123 @@
...
+
+    for (auto hit : sortedHits){
```
wkrzemien: probably should be 
for (const auto & hit : sortedHits)
{
}

Alvarness:
HitFinderTools as an independent module (for now). What is a "clean" (recommended) method for using typedefs like SignalsContainer?

wkrzemien:
Well syntactically you can use:
typedef std::vector<MyObject> MyName
or
using MyName = std::vector<MyObject> (from c++11)
I don't have any preferences. I only tend to capitalize the first letter. And I rather define it in the class scope.
If you were asking about the semantic and naming convention. Then I think it is a question of taste.
The main idea I think is to make code more expressive, meaningful and shorter.

LargeBarrel/HitFinder.cpp
```diff
@@ -0,0 +1,122 @@
...
+                saveHits(createHits( fAllSignalsInTimeWindow, kTimeWindowWidth));
+                fAllSignalsInTimeWindow.clear();
+                fillSignalsMap(*currSignal);
+            }
+        }
+    }
+}
+
+
+vector<JPetHit> HitFinder::createHits(const SignalsContainer& allSignalsInTimeWindow, const double timeDifferenceWindow){
```
wkrzemien: Just use the  createHits from HitFinderTools , and remove createHits from HitFinder

LargeBarrel/HitFinderTools.cpp
```diff
@@ -0,0 +1,54 @@
+/**
```
wkrzemien: This seems perfectly fine. Now add HitFinderToolsTests.cpp with tests

Alvarness:
First test is working. Next three however causes segmentation violation - don't know why

PR #21, opened 2017-01-07
- added KrzysztofAnalysis

kkacprzak:
first version of signal matching


KrzysztofAnalysis/Module2.h
```diff
@@ -0,0 +1,58 @@
+/**
```
wkrzemien: Nice work.
Change the name of the Module2 to e.g. SignalFinder. (per analogy Nikodem used the name HitFinder for his module).

KrzysztofAnalysis/Module2.h
```diff
@@ -0,0 +1,58 @@
...
+	protected:
+		JPetWriter* fWriter;
+		JPetParamManager* fParamManager;
+		LargeBarrelMapping fBarrelMap;
+		void saveRawSignals(vector<JPetRawSignal> sigChVec);
+		vector<JPetRawSignal> buildRawSignals(Int_t timeWindowIndex, vector<JPetSigCh> sigChFromSamePM);
+		int findSigChOnNextThr(JPetSigCh sigCh, vector<JPetSigCh> sigChVec);
+		int findTrailingSigCh(JPetSigCh leadingSigCh, vector<JPetSigCh> trailingSigChVec);
+		const int kNumOfThresholds = 4;
+		const Float_t SIGCH_EDGE_MAX_TIME = 20000; //[ps]
```
wkrzemien: we have convention (adopted) to call const as kMyConst, so please change it to float kSigCHEdgeMaxTime and the next one too.

KrzysztofAnalysis/Module2.h
```diff
@@ -0,0 +1,58 @@
...
+		const Float_t SIGCH_LEAD_TRAIL_MAX_TIME = 300000; //[ps]
```
wkrzemien: the same

KrzysztofAnalysis/Module2.h
```diff
@@ -0,0 +1,58 @@
...
+
+#include <vector>
+#include <JPetTask/JPetTask.h>
+#include <JPetRawSignal/JPetRawSignal.h>
+#include <JPetTimeWindow/JPetTimeWindow.h>
+#include <JPetParamBank/JPetParamBank.h>
+#include <JPetParamManager/JPetParamManager.h>
+#include "LargeBarrelMapping.h"
+
+using namespace std;
```
wkrzemien: using namespace in a header file is a bad practice. It leads to namespace pollution cause every file that includes your header gets automatically included all things from this namespace.
So using namespace std; only in cpp.
Please change it.

KrzysztofAnalysis/Module2.cpp
```diff
@@ -0,0 +1,251 @@
...
+ *  you may not use this file except in compliance with the License.
+ *  You may find a copy of the License in the LICENCE file.
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ *  @file TaskB1.cpp
```
wkrzemien: change it

KrzysztofAnalysis/Module2.cpp
```diff
@@ -0,0 +1,251 @@
...
+}
+
+//Module execution method
+void Module2::exec(){
+
+	//getting the data from event in propriate format
+	if(auto timeWindow = dynamic_cast<const JPetTimeWindow*const>(getEvent())){
+
+		map<int,vector<JPetSigCh>> sigChsPMMap;
+		//map<int,vector<JPetRawSignal>> rawSignalsPMMap;
```
wkrzemien: remove commented code

KrzysztofAnalysis/Module2.cpp
```diff
@@ -0,0 +1,251 @@
...
+//Module execution method
+void Module2::exec(){
+
+	//getting the data from event in propriate format
+	if(auto timeWindow = dynamic_cast<const JPetTimeWindow*const>(getEvent())){
+
+		map<int,vector<JPetSigCh>> sigChsPMMap;
+		//map<int,vector<JPetRawSignal>> rawSignalsPMMap; 
+
+		//map Signal Channels in this Time window according to PM they belong to
```
wkrzemien: remove

KrzysztofAnalysis/Module2.h
```diff
@@ -0,0 +1,58 @@
...
+		virtual void terminate() override;
+		virtual void setWriter(JPetWriter* writer) override;
+		virtual void setParamManager(JPetParamManager* paramManager) override;
+		const JPetParamBank& getParamBank()const;
+
+	protected:
+		JPetWriter* fWriter;
+		JPetParamManager* fParamManager;
+		LargeBarrelMapping fBarrelMap;
+		void saveRawSignals(vector<JPetRawSignal> sigChVec);
```
wkrzemien: Probably for all the methods that use vectors as arguments you should change it to const references:
e.g.
void saveRawSignals(vector<JPetRawSignal> sigChVec);
to void saveRawSignals(const vector<JPetRawSignal> & sigChVec);
or
vector<JPetRawSignal> buildRawSignals(Int_t timeWindowIndex, vector<JPetSigCh> sigChFromSamePM) 
to
vector<JPetRawSignal> buildRawSignals(Int_t timeWindowIndex, const vector<JPetSigCh>& sigChFromSamePM) 
and so on

Passing by reference avoids the copy of the whole vector, const assures that you dont change the input parameter.

KrzysztofAnalysis/Module2.cpp
```diff
@@ -0,0 +1,251 @@
...
+				thresholdSigCh.at(7).erase(thresholdSigCh.at(7).begin()+closestTrailingSigCh3);
+			}
+
+			rawSig->addPoint(thresholdSigCh.at(3).at(nextThrSigChIndex3));
+			thresholdSigCh.at(3).erase(thresholdSigCh.at(3).begin()+nextThrSigChIndex3);
+		}
+
+		//addingcreated Raw Signal to vector
+		rawSigVec.push_back(*rawSig);
+		thresholdSigCh.at(0).erase(thresholdSigCh.at(0).begin()+0);
```
wkrzemien: what is that?

KrzysztofAnalysis/Module2.cpp
```diff
@@ -0,0 +1,251 @@
...
+	getStatistics().getHisto1D("remainig_trailing_sig_ch_per_thr").Fill(3,thresholdSigCh.at(6).size());
+	getStatistics().getHisto1D("remainig_trailing_sig_ch_per_thr").Fill(4,thresholdSigCh.at(7).size());
+	
+	//return argument
+	return rawSigVec;
+}
+
+//method of finding Signal Channels that belong to the same leading edge 
+//not more than some amount of ps away, defined in header file
+int Module2::findSigChOnNextThr(JPetSigCh sigCh, vector<JPetSigCh> sigChVec){
```
wkrzemien: for tests later I woudl explicitely add time window argument so instead of
int Module2::findSigChOnNextThr(JPetSigCh sigCh, vector<JPetSigCh> sigChVec)
do 
int Module2::findSigChOnNextThr(const JPetSigCh& sigCh, const vector<JPetSigCh>& sigChVec, int sigChEdgeMaxTime)
{
...
}
Or even you dont need to pass sigCh but only double  taken from  sigCh.getValue()?

KrzysztofAnalysis/Module2.cpp
```diff
@@ -0,0 +1,251 @@
...
+	}
+	if(trailingFoundIdices.size()==0) return -1;
+	sort(trailingFoundIdices.begin(), trailingFoundIdices.end());
+	return trailingFoundIdices.at(0);
+}
+
+//saving method
+void Module2::saveRawSignals(vector<JPetRawSignal> sigChVec){
+	assert(fWriter);
+	for(JPetRawSignal sigCh : sigChVec){
```
wkrzemien: for(const auto& sigCh : sigChVec)

KrzysztofAnalysis/Module2.cpp
```diff
@@ -0,0 +1,251 @@
...
+
+		//fourth thr leading and trailing
+		int nextThrSigChIndex3 = findSigChOnNextThr(thresholdSigCh.at(0).at(0), thresholdSigCh.at(3));
+		if(nextThrSigChIndex3 != -1) {
+
+			int closestTrailingSigCh3 = findTrailingSigCh(thresholdSigCh.at(3).at(nextThrSigChIndex3), thresholdSigCh.at(7));
+			if(closestTrailingSigCh3 != -1) {
+				double tot3 = thresholdSigCh.at(7).at(closestTrailingSigCh3).getValue() 
+						- thresholdSigCh.at(3).at(nextThrSigChIndex3).getValue();
+				getStatistics().getHisto1D("TOT_thr_4").Fill(tot3/1000.0);
```
wkrzemien: In the final version I would remove all the getStatistics or
define some bool fSaveControlHistos; in the class definition and put it everywhere to be able to "switch off" histograms if not needed
e.g
if (fSaveControlHistos) {
double tot3 = thresholdSigCh.at(7).at(closestTrailingSigCh3).getValue() 
 +						- thresholdSigCh.at(3).at(nextThrSigChIndex3).getValue();
 +				getStatistics().getHisto1D("TOT_thr_4").Fill(tot3/1000.0);
}

KrzysztofAnalysis/Module2.cpp
```diff
@@ -0,0 +1,251 @@
...
+
+Module2::Module2(const char * name, const char * description) : JPetTask(name, description){}
+
+Module2::~Module2(){}
+
+//Module init method
+void Module2::init(const JPetTaskInterface::Options&){
+	INFO("Module 2 started.");
+	fBarrelMap.buildMappings(getParamBank());
+	getStatistics().createHistogram(new TH1F("remainig_leading_sig_ch_per_thr", "Remainig Leading Signal Channels", 4, 0.5, 4.5));
```
wkrzemien: all those guys should be or removed or activated conditionally - see later comments

KrzysztofAnalysis/Module2.cpp
```diff
@@ -0,0 +1,251 @@
...
+void Module2::exec(){
+
+	//getting the data from event in propriate format
+	if(auto timeWindow = dynamic_cast<const JPetTimeWindow*const>(getEvent())){
+
+		map<int,vector<JPetSigCh>> sigChsPMMap;
+		//map<int,vector<JPetRawSignal>> rawSignalsPMMap; 
+
+		//map Signal Channels in this Time window according to PM they belong to
+		const unsigned int nSigChs = timeWindow->getNumberOfSigCh();
```
wkrzemien: The lines from 46 to 62 I would extract to a function e.g.

std::map<int, std::vector<JPetSigCh>> getSigChByPMId(const JPetTimewWindow* window) const
{
...
return sigChsPMMap;
}

KrzysztofAnalysis/Module2.cpp
```diff
@@ -0,0 +1,251 @@
...
+				vector<JPetSigCh> tmp;
+				tmp.push_back(sigCh);
+				sigChsPMMap.insert(pair<int,vector<JPetSigCh>>(pmt_id,tmp));
+			}else{
+				search->second.push_back(sigCh);
+			}
+		}
+
+		//reading Signal Channel Map and creating Raw Signals
+		for(auto &sigChPair : sigChsPMMap){
```
wkrzemien: Next, I would extract another function here (instead of 64 to 69:
const std::vector<JPetRawSignals>&  buildSignals(const std::map<int, std::vector<SigCh>>& sgiChsPMMap, int timeWindowIndex) 
{
std::vector<JPetRawSignal> allSignals;
for(auto &sigChPair : sigChsPMMap) {
std::vector<JPetRawSignal> currentSignals = buildRawSignals(timeWindow->getIndex(), sigChPair.second);
allSignals.insert(allSignals.end(), currentSignals.begin(), currentSignals.end());
}
/// I am not sure if it will be optimized out
return allSignals;
}

and here I would just call:


And here I would just call:
std::vector<JPetRawSignal> allSignals = buildSignals(sgiChsPMMap, timeWindow->getIndex());
saveRawSignals(allSignals);

Just the idea. Probably can be done better.

KrzysztofAnalysis/Module2.cpp
```diff
@@ -0,0 +1,251 @@
...
+//sorting method
+bool sortByTimeValue(JPetSigCh sig1, JPetSigCh sig2){
+	return (sig1.getValue() < sig2.getValue());
+}
+
+//method creating Raw signals form vector of Signal Channels
+vector<JPetRawSignal> Module2::buildRawSignals(Int_t timeWindowIndex, vector<JPetSigCh> sigChFromSamePM){
+	
+	vector<JPetSigCh> tmp;
+	vector<vector<JPetSigCh>> thresholdSigCh;
```
wkrzemien: maybe it would be easier to use 
`std::map<int , std::vector<JPetSigCh>>= {{0,tmp}, {1,tmp} ....};` where first int is the threshold

KrzysztofAnalysis/Module2.cpp
```diff
@@ -0,0 +1,251 @@
...
+
+	//initialisation with empty vectors
+	for(int i=0;i<2*kNumOfThresholds;++i){
+		thresholdSigCh.push_back(tmp);
+
+	}
+
+	//division into subvectors according to threshold number:
+	//0-3 leading, 4-7 trailing
+	for(JPetSigCh sigCh : sigChFromSamePM){
```
wkrzemien: so if you need to divide it to leading and trailing it is better to do it already while creating sigChFromSamePM
so sigChFromSamePM can be instead of std::vector<SigCh> 
e.g std::pair<enum of Trailing and Leading, std::vector<SigCh>>

KrzysztofAnalysis/Module2.cpp
```diff
@@ -0,0 +1,251 @@
...
+				getStatistics().getHisto1D("TOT_thr_3").Fill(tot2/1000.0);
+				rawSig->addPoint(thresholdSigCh.at(6).at(closestTrailingSigCh2));
+				thresholdSigCh.at(6).erase(thresholdSigCh.at(6).begin()+closestTrailingSigCh2);
+			}
+
+			rawSig->addPoint(thresholdSigCh.at(2).at(nextThrSigChIndex2));
+			thresholdSigCh.at(2).erase(thresholdSigCh.at(2).begin()+nextThrSigChIndex2);
+		}
+
+		//fourth thr leading and trailing
```
wkrzemien: There is some redundancy in this code that for sure can be removed, I will try to think about it.

KrzysztofAnalysis/Module2.cpp
```diff
@@ -0,0 +1,251 @@
...
+void Module2::terminate(){
+	INFO("Module 2 ended.");
+}
+
+//sorting method
+bool sortByTimeValue(JPetSigCh sig1, JPetSigCh sig2){
+	return (sig1.getValue() < sig2.getValue());
+}
+
+//method creating Raw signals form vector of Signal Channels
```
wkrzemien: Since this is the heart of your algorithm please add some description here

KrzysztofAnalysis/Module2.cpp
kkacprzak: removing first entry from vector in while loop. Main idea is to find raw signals for every first threshold on leading edge SigCh, when signal is constructed, there is no need to hold it in the vector, so I am removing it. Later, histogram with remaining SigCh can be drawn, and threshold should be empty.

KrzysztofAnalysis/Module2.cpp
kkacprzak: yes, that would be helpful.

KrzysztofAnalysis/Module2.h
kkacprzak: done

KrzysztofAnalysis/Module2.h
kkacprzak: done

KrzysztofAnalysis/Module2.h
kkacprzak: done

KrzysztofAnalysis/Module2.h
kkacprzak: done

KrzysztofAnalysis/Module2.cpp
kkacprzak: done

KrzysztofAnalysis/Module2.cpp
kkacprzak: done

KrzysztofAnalysis/Module2.h
kkacprzak: done

KrzysztofAnalysis/Module2.cpp
kkacprzak: done

KrzysztofAnalysis/Module2.cpp
kkacprzak: TODO

KrzysztofAnalysis/Module2.cpp
kkacprzak: TODO

KrzysztofAnalysis/Module2.cpp
kkacprzak: done

KrzysztofAnalysis/Module2.cpp
kkacprzak: done

KrzysztofAnalysis/Module2.cpp
kkacprzak: done

KrzysztofAnalysis/Module2.cpp
kkacprzak: done

KrzysztofAnalysis/Module2.cpp
kkacprzak: done

PR #25, opened 2017-02-02
[WIP] Large barrel extended example

wkrzemien:
Large Barrel Extended Example - including time calibration loading, Signal and Hit Finding.
Version under test


wkrzemien:
Large Barrel Extended Example - including time calibration loading, Signal and Hit Finding.
Version under test

PR #28, opened 2017-03-07
Extension of Extended Exapmle

kkacprzak:
Extension of HitFinder, EventFinder and additin of EventCategorizer. Also necessary files for calibration were added.


wkrzemien:
this one is outdated

PR #29, opened 2017-03-16
LargeBarrelExtended

wkrzemien:
New example of the LargeBarrel Analysis including:
1) time calibration loading
2) signal matching
3) hit matching
4)event matching


LargeBarrelAnalysisExtended/conf_trb3.xml
```diff
@@ -0,0 +1,353 @@
+<?xml version="1.0"?>
```
alekgajos: This file should not be in the repo.

LargeBarrelAnalysisExtended/run.sh
```diff
@@ -0,0 +1 @@
+./LargeBarrelAnalysisExtended.x -t hld -f <data file name>.hld -p conf_trb3.xml -u userParams.json -i 43 -l large_barrel.json
```
alekgajos: With <data file name> this is not a valid shell script, it would better fit in a readme. Otherwise, the filename could be made a parameter of the script.

wkrzemien: done

LargeBarrelAnalysisExtended/conf_trb3.xml
wkrzemien: removed

PR #35, opened 2017-04-30
Copy files only if they don't exist

grey277:


wkrzemien:
Please review @alekgajos

PR #36, opened 2017-05-04
Reco output

wictus:
Prepared new example with output ready for reconstruction library.
The output file has to be converted further by program from reconstruction library to match internal format.
**Calibration files are added to this PR but if they are not need let me know.**


alekgajos:
Do you perhaps have a version of this module adapted to Framework version 6?
Or could you adapt it? We could then incorporate it in the next release.

wictus:
I can adapt it.

PR #38, opened 2017-07-05
Smarter data downloading

timorl:
Necessary changes at sphinx already done, everything should work better than previously.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -25,12 +25,9 @@ if(APPLE)
 endif()
 
 ##download input files and config files
-#only once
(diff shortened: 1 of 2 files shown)


wkrzemien:
Meanwhile there was a one pull request which slightly changed download_data.sh
basically add a line to download additional folder. Could you remove the confiict?

timorl:
Done. The only thing i needed to do was undoing the conflicting commit, because with my changes the CalibrationFiles are already being downloaded, and specifying them explicitly in the script is unnecessary. The checksums file on sphinx was already updated, because by the time I was improving the download script they were already present there.

PR #41, opened 2017-08-08
Fix z axis position

wictus:
Z axis position is now correctly calculated.
Tests for HitFinderTools checking if Z axis convention is held are also added.

### DIFF
--- a/LargeBarrelAnalysisExtended/HitFinderToolsTest.cpp
+++ b/LargeBarrelAnalysisExtended/HitFinderToolsTest.cpp
@@ -16,23 +16,48 @@ BOOST_AUTO_TEST_SUITE(FirstSuite)
 BOOST_AUTO_TEST_CASE( checkZAxisConvention )
 { 
     JPetStatistics stats;
-    SignalsContainer allSignalsInTimeWindow;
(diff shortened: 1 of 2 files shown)


LargeBarrelAnalysisExtended/HitFinderToolsTest.cpp
```diff
@@ -0,0 +1,63 @@
...
+
+typedef std::map <int, std::pair < std::vector<JPetPhysSignal>, std::vector<JPetPhysSignal> > > SignalsContainer;
+
+
+BOOST_AUTO_TEST_SUITE(FirstSuite)
+
+BOOST_AUTO_TEST_CASE( checkZAxisConvention )
+{ 
+    JPetStatistics stats;
+    stats.createHistogram( new TH2F("time_diff_per_scin", "time_diff_per_scin" ,100,0,100, 100, 0, 100));
```
wkrzemien: Do you really need to create those histograms?

wictus: Unfortunately yes. In HitFinderTools.cpp line 88 those histos are filled by createHits function.

PR #56, opened 2017-12-11
Add test to check is barrel slot's theta is degree or radian

grey277:
Refactory of HitFinderTools, split it into separate functions for easier testing.
Also I added test, that checks is there is barrel slot with theta greater then 7(radians should be from 0 to 2PI)


grey277:
Added fix for issue 909, also some refactoring and optimization. On 10000 events there is about 10% increase on speed.
New user option: HitFinder_Velocity_Calibration_File_Path_std::string with path to velocity calibration file

LargeBarrelAnalysis/HitFinderTools.cpp
```diff
@@ -99,43 +143,34 @@ vector<JPetHit> HitFinderTools::createHits(JPetStatistics& stats,
...
+{
+  bool isDegree = false;
+  for (const JPetHit& hit : hits) {
+    if (hit.getBarrelSlot().getTheta() > 7.) {
+      isDegree = true;
+      break;
+    }
+  }
+  return isDegree;
+}
```
wkrzemien: end of line missing

wkrzemien:
Also, change j-pet-framework to latest develop version

PR #66, opened 2018-02-09
Fix to velocity calibration master

wictus:
Fix of velocity calibration example, now it works on the data.
UserParams file is filled with "dummy" paths to not allow blind analysis


VelocityCalibration/DeltaTFinder.cpp
```diff
@@ -105,12 +107,19 @@ const char* DeltaTFinder::formatUniqueSlotDescription(const JPetBarrelSlot & slo
...
+    uint nhits = timeWindow->getNumberOfEvents();
+    for (uint i = 0; i < nhits; ++i) {
+        fillHistosForHit( dynamic_cast<const JPetHit&>(timeWindow->operator[](i)) );
+    }
+        }
+    else{
+      return false;
+        }
+
+
```
alekgajos: Please fix the identation of brakets.

wictus: Can you check it now?

PR #68, opened 2018-02-18
Fix to velocity calibration develop

wictus:
Similiar merge to the one on master branch


alekgajos:
Shouldn't thir PR be targetting master instead of develop?

wictus:
There was one targetting master. This one is for develop, as requested by WK.

PR #69, opened 2018-02-19
Add module creating sinograms from JPetHits

grey277:
Adding module creating sinogram from JPetHits and also moves ImageReco module to new folder: ImageReconstruction.

It is using data from all layers and defines 5 user options:
 - "SinogramCreator_OutFileName_std::string": defines output file name where sinogram is saved
 - "SinogramCreator_ReconstructionLayerRadius_float": defines radius of reconstruction layer
 - "SinogramCreator_ReconstructionDistanceAccuracy_float": defines maximal round value for distance, in cm, e.g. 0.1 means 1 bin in sinogram corresponds to 0.1 cm in reality
 - "SinogramCreator_SinogramZSplitNumber_int": defines number of splits around "z" coordinate
 - "SinogramCreator_ScintillatorLenght_float": defines scintillator lenght in "z" coordinate


ImageReconstruction/CMakeLists.txt
```diff
@@ -0,0 +1,57 @@
+# J-PET tool project using J-PET framework
+#
+# Author: Adam Strzelecki <adam.strzelecki@uj.edu.pl>
+#
+# Description:
+#   Builds project depending on J-PET framework.
+
+cmake_minimum_required(VERSION 2.6)
+
+######################################################################
```
alekgajos: You can remove this comment as it is only relevant in the template.

ImageReconstruction/SinogramCreator.cpp
```diff
@@ -0,0 +1,134 @@
...
+    }
+  } else {
+    ERROR("Returned event is not TimeWindow");
+    return false;
+  }
+  return true;
+}
+
+bool SinogramCreator::checkLayer(const JPetHit& hit)
+{
```
alekgajos: Perhaps it would be good to add a comment here with a comprehensive explanation of this requirement.

ImageReconstruction/SinogramCreator.h
```diff
@@ -0,0 +1,81 @@
...
+  const std::string kReconstructionLayerRadiusKey = "SinogramCreator_ReconstructionLayerRadius_float";
+  const std::string kReconstructionStartAngle = "SinogramCreator_ReconstructionStartAngle_float";
+  const std::string kReconstructionEndAngle = "SinogramCreator_ReconstructionEndAngle_float";
+  const std::string kReconstructionDistanceAccuracy = "SinogramCreator_ReconstructionDistanceAccuracy_float";
+  const std::string kReconstructionAngleStep = "SinogramCreator_ReconstructionAngleStep_float";
+
+  std::string fOutFileName = "sinogram.ppm";
+  unsigned int fMaxValueInSinogram = 0; // to fill later in output file
+
+  float fReconstructionLayerRadius = 42.5f;
```
alekgajos: It would be better to extract this radius from the ParamBank rather than to have it hard-coded.

ImageReconstruction/SinogramCreatorToolsTest.cpp
```diff
@@ -0,0 +1,168 @@
...
+  SinogramResultType* fSinogram = nullptr;
+  unsigned int maxThetaNumber = std::ceil(180. / fReconstructionAngleStep);
+  unsigned int maxDistanceNumber = std::ceil(fReconstructionLayerRadius * 2 * (1.f / fReconstructionDistanceAccuracy)) + 1;
+  if (fSinogram == nullptr) {
+    fSinogram = new SinogramResultType(maxDistanceNumber, (std::vector<unsigned int>(maxThetaNumber)));
+  }
+  unsigned int currentValueInSinogram = 0;
+  unsigned int fMaxValueInSinogram = 0;
+  std::cout << "Number of hits: " << hits.size() << std::endl;
+  std::cout << "Hit[0]: x:" << hits[0].first.first << " y:" << hits[0].first.second << " x:" << hits[0].second.first << " y:" << hits[0].second.second << std::endl;
```
alekgajos: Please remove the stray cout-s.

alekgajos:
I have only one more concern - now the user can provide the radius of reconstruction layer, which in principle can be different from the actual radii of the layers in the setup?
Maybe it would be better if the user only specified tha layer number and its radius was extracted from ParamBank? Or at least a warning should be logged if the user-provided radius does not match any of the layers present in the ParamBank.

PR #76, opened 2018-04-23
Fix test for radian/degree barrel slot theta angle

grey277:
Fix test for radian/degree barrel slot theta angle


wkrzemien:
Please solve the conflicts

PR #77, opened 2018-05-14
Reco output f6

wictus:
Output for MLEM reconstruction. Shows in what format one should extract data. This version is for framework 6 (master). The previous PR should be deleted.


wkrzemien:
Kamil R. is going to add it separately

PR #79, opened 2018-05-31
Fix of const values in EvCatTool

kkacprzak:
Described in here
[http://sphinx.if.uj.edu.pl/redmine/issues/955](Redmine)


kkacprzak:
It passes test with gcc 4.8, boost 1.58 and root 5.34 with travis and jenkins. 
With those small changes also it is also possible to build with gcc 7.3, boost 1.65 and root 6.10/08

PR #80, opened 2018-05-31
MCGeantAnalysis

daria137:
- MCGeantAnalysis version adjusted for current develop version (root6 working :) 
- MCGeantAnalysis - run.sh allows to process both data or mcGeant files
- MCGeantAnalysis is strongly based on LargeBarrelAnalysis (new calibration files required)
- (ugly) config files are copies; final solution should be based on folder 'CalibrationFiles'
- MC files are not included in the repository


MCGeantAnalysis/CMakeLists.txt
```diff
@@ -0,0 +1,72 @@
...
+
+######################################################################
+### when creating a new project,
+### set this section appropriately for your project
+######################################################################
+set(projectName MCGeantAnalysis)
+
+set(AUXILLIARY_FILES
+  README.md
+  run.sh
```
alekgajos: The auxilliary files should be copied from the common location. This, however, has to be done after issue 957 (at redmine) is solved.

MCGeantAnalysis/main.cpp
```diff
@@ -0,0 +1,79 @@
+/**
+ *  @copyright Copyright 2017 The J-PET Framework Authors. All rights reserved.
```
alekgajos: 2018 :-)

MCGeantAnalysis/main.cpp
```diff
@@ -0,0 +1,79 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ *  @file main.cpp
+ */
+
+#include <JPetManager/JPetManager.h>
+#include <JPetGeantParser/JPetGeantParser.h>
+
+#include "TimeWindowCreator.h"
```
alekgajos: These files should be included from LargeBarrelAnalysis unless a custom version of some module is required.

MCGeantAnalysis/main.cpp
```diff
@@ -0,0 +1,79 @@
...
+#include "./JPetOptionsTools/JPetOptionsTools.h"
+#include "./JPetCmdParser/JPetCmdParser.h" 
+#include "./JPetOptionsGenerator/JPetOptionsGenerator.h"
+
+using namespace std;
+
+int main(int argc, const char* argv[])
+{
+
+    JPetOptionsGenerator optionsGenerator;
```
alekgajos: Question to @wkrzemien: is this the only possible way to access an option from within main()?
I understand the same parsing is done internally in the Manager, so it should be better to avoid repeating it manually here.

MCGeantAnalysis/main.cpp
```diff
@@ -0,0 +1,79 @@
...
+    JPetCmdParser parser;
+    auto optionsFromCmdLine = parser.parseCmdLineArgs(argc, argv); 
+    std::map<std::string, boost::any> fOptions = optionsGenerator.generateAndValidateOptions(optionsFromCmdLine); 
+    auto fileType = jpet_options_tools::FileTypeChecker::getInputFileType(fOptions);
+
+
+    JPetManager& manager = JPetManager::getManager();
+
+    if (fileType == jpet_options_tools::FileTypeChecker::kMCGeant) {
```
alekgajos: I believe JPetGeantParser should be an "internal" module of the same kind as HLDreader of ScopeLoader, so that it would be automatically added to the beginning of the processing chain if the file type is kMCGeant.
This if should not be necessary here then.

MCGeantAnalysis/main.cpp
alekgajos: Ok, if we move the JPetGeantParser to work the same way as Unpacker does, this code will not be needed in main anyway.

wkrzemien: Yes, that could be the solution.  I am also keen on not doing it manually.

MCGeantAnalysis/main.cpp
wkrzemien: Yeap

alekgajos:
This PR was replaced with #87 for easier maintenance.

PR #81, opened 2018-06-04
Aux files cleanup

alekgajos:
Updating the core repo to the version with universal CMake.
Making all of the examples copy auxilliary files from their own source directories rather than from LargeBarrelAnalysis. (fixes: #951 at Redmine)


alekgajos:
Just as said in the PR description:-)
This is intenitonal. It is not required by the changes introduced to the examples, except that it allows to run the tests with ROOT 6 as well. However, if not in this PR, we should update the core submodule in a separate one as soon as possible.

wkrzemien:
Sorry, my bad :)

PR #82, opened 2018-06-04
Extract FilterEvents UserTask, refactory and fixes to SinogramCreator

grey277:
It is extension for #69.
I extracted FilterEvents JPetUserTask, because multiple image reconstruction algorithms  can use same filtered events.

- Fix forgotten initialize to max sinogram value 
- Add skip for 90 degree(there is some bug that needs to be fixed). 
- Add swapping first/second point, to keep good distance sign 
- Swap y axis sign to match tomograph axes.
- Change Dockerfile to build root 6 with CMake
- Fix saving reconstruction with pixel value < 0


ImageReconstruction/FilterEvents.cpp
```diff
@@ -0,0 +1,174 @@
+
```
wkrzemien: remove first line

ImageReconstruction/FilterEvents.h
```diff
@@ -0,0 +1,75 @@
...
+
+  const std::string kCutOnTOTMinValueKey = "FilterEvents_TOT_MIN_VALUE_IN_NS_float";
+  const std::string kCutOnTOTMaxValueKey = "FilterEvents_TOT_MAX_VALUE_IN_NS_float";
+  const std::string kCutOnAngleDeltaMinValueKey = "FilterEvents_ANGLE_DELTA_MIN_VALUE_float";
+
+  const int kNumberOfHitsInEventHisto = 10;
+  const int kNumberOfConditions = 6;
+
+  float fCUT_ON_Z_VALUE = 23;
+  float fCUT_ON_LOR_DISTANCE_FROM_CENTER = 25;
```
wkrzemien: can you add  units after // 
It will be easier for the reader

ImageReconstruction/ImageReco.cpp
```diff
@@ -269,22 +111,67 @@ bool ImageReco::calculateAnnihilationPoint(const JPetHit& firstHit, const JPetHi
...
+  {
+    fANNIHILATION_POINT_Z = getOptionAsFloat(opts, kCutOnAnnihilationPointZKey);
+  }
+
+  if (isOptionSet(opts, kBinMultiplierKey))
+  {
+    fBinMultiplier = getOptionAsDouble(opts, kBinMultiplierKey);
+    if ((std::floor(fBinMultiplier * fXRange) *
+         std::floor(fBinMultiplier * fYRange) *
+         std::floor(fBinMultiplier * fZRange)) > 1073741822)
```
wkrzemien: Please change this magic number to some meanigful constant

ImageReconstruction/ImageReco.cpp
```diff
@@ -269,22 +111,67 @@ bool ImageReco::calculateAnnihilationPoint(const JPetHit& firstHit, const JPetHi
...
+  {
+    fZRange = getOptionAsInt(opts, kZRangeOn3DHistogramKey);
+  }
+
+  if (isOptionSet(opts, kCutOnAnnihilationPointZKey))
+  {
+    fANNIHILATION_POINT_Z = getOptionAsFloat(opts, kCutOnAnnihilationPointZKey);
+  }
+
+  if (isOptionSet(opts, kBinMultiplierKey))
```
wkrzemien: Some comment about this piece of code would be useful.

modules/large_barrel/JPetRecoImageTools/JPetRecoTest.cpp
```diff
@@ -107,7 +111,11 @@ BOOST_AUTO_TEST_CASE(backProjectSinogramRamLak)
   res << maxValue << std::endl;
   for (unsigned int i = 0; i < result.size(); i++) {
     for (unsigned int j = 0; j < result[0].size(); j++) {
-      res << static_cast<int>(result[i][j]) << " ";
+      int resultInt = static_cast<int>(result[i][j]);
```
wkrzemien: I know those are tests, but this double loop is repeated 3 times, mabye it would be better to extract it to some helper function within tests

wkrzemien: Actually 5 times

ImageReconstruction/SinogramCreatorTools.cpp
```diff
@@ -52,5 +55,10 @@ float SinogramCreatorTools::calculateDistance(float firstX, float firstY, float
 
 float SinogramCreatorTools::calculateNorm(float firstX, float firstY, float secondX, float secondY)
 {
+  if (firstX > secondX)
+  {
+    std::swap(firstX, secondX);
+    std::swap(firstY, secondY);
+  }
   return std::sqrt(std::pow((secondY - firstY), 2) + std::pow((secondX - firstX), 2));
 }
```
wkrzemien: end of line missing

ImageReconstruction/SinogramCreatorTools.cpp
```diff
@@ -52,5 +55,10 @@ float SinogramCreatorTools::calculateDistance(float firstX, float firstY, float
 
 float SinogramCreatorTools::calculateNorm(float firstX, float firstY, float secondX, float secondY)
 {
+  if (firstX > secondX)
```
wkrzemien: why do you need to swap it?

grey277: Swapping is needed to ensure same distance sign with swapped arguments, later on I'm checking angle and if angle > 90 I'm changing distance sign to negative.

wkrzemien:
Please, add also some type of general purpose README or extended comment describing it Since it will be used by people we should  be sure that it it documented.

grey277:
I extended more describing commend, is it enough? If not, what more should this comment contains?

PR #89, opened 2018-06-30
Redmine tasks 957 & 958

kkacprzak:
Removal of obsolete LargeBarrelMapping and small config files handling. I updated ExamplesReformed directory on sphinx, so now all the files are downloaded to new directory 'CalibrationFiles' inside source folder.


LargeBarrelAnalysis/CMakeLists.txt
```diff
@@ -14,7 +14,9 @@ cmake_minimum_required(VERSION 2.6)
 set(projectName LargeBarrelAnalysis)
 
 set(AUXILIARY_FILES
-  run.sh
+  conf_trb3.xml
```
wkrzemien: Are those files should be there, or it was added by accident?

alekgajos: I think we should remove conf_trb3.xml from this list. Right now there is only one version but this may not be true in the future (e.g. after 4th layer is added) so I would leave it in the CalibrationFiles directory for the users to choose.

PR #92, opened 2018-07-12
Iterative example of Time Calibration

wkrzemien:
Add iterative version of Time Calibration based on the new iterative scheme.
This approach uses several new mechanisms which are included into the dev core framework


alekgajos:
Superseded by #126.

PR #87, opened 2018-06-19
MC wrapper (taken over from daria137)

alekgajos:
This was originally PR #80 from @daria137 
I have copied it and created a new PR to temporarily take over maintenance of this code.

PR #93, opened 2018-07-16
MC wrapper PR after a cleanup (replaces PR #87)

alekgajos:
This PR may only be accepted after PR #181 to core framework.

Redundant files were removed and the example is now as simple as possible.


wkrzemien:
Actually, there is a conflict to solve

PR #94, opened 2018-07-26
Update submodules main.cpp to new iterative version

grey277:
Add try .. catch to main.cpp, update submodule to newest develop version

### DIFF
--- a/TimeCalibration/main.cpp
+++ b/TimeCalibration/main.cpp
@@ -13,33 +13,36 @@
  *  @file main.cpp
  */
 
-#include <JPetManager/JPetManager.h>
-#include "../LargeBarrelAnalysis/TimeWindowCreator.h"
+#include "../LargeBarrelAnalysis/EventCategorizer.h"
+#include "../LargeBarrelAnalysis/EventFinder.h"
+#include "../LargeBarrelAnalysis/HitFinder.h"
 #include "../LargeBarrelAnalysis/SignalFinder.h"
 #include "../LargeBarrelAnalysis/SignalTransformer.h"
-#include "../LargeBarrelAnalysis/HitFinder.h"
-#include "../LargeBarrelAnalysis/EventFinder.h"
-#include "../LargeBarrelAnalysis/EventCategorizer.h"
+#include "../LargeBarrelAnalysis/TimeWindowCreator.h"
 #include "TimeCalibration.h"
+#include <JPetManager/JPetManager.h>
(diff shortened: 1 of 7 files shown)


wkrzemien:
@alekgajos  Can you have a look what is wrong here?

TimeCalibration/main.cpp
```diff
@@ -13,33 +13,36 @@
...
+int main(int argc, const char* argv[]) {
+  try {
+    JPetManager& manager = JPetManager::getManager();
 
-  manager.useTask("TimeWindowCreator", "hld", "tslot.calib");
-  manager.useTask("SignalFinder", "tslot.calib", "raw.sig");
-  manager.useTask("SignalTransformer", "raw.sig", "phys.sig");
-  manager.useTask("HitFinder", "phys.sig", "hits");
-  manager.useTask("TimeCalibration", "hits", "calib");
+    manager.registerTask<TimeWindowCreator>("TimeWindowCreator");
```
wkrzemien: is the formatting ok?

grey277:
The formatting seems ok, I added try/catch scope so it added extra tab.
The problem are different/not existing user tasks, in main.cpp of LargeBarellAnalysis.x there is:

    manager.useTask("TimeWindowCreator", "hld", "tslot.calib");
    manager.useTask("SignalFinder", "tslot.calib", "raw.sig");

but test is doing:
sed -i 's/manager.useTask("TimeWindowCreator", "hld", "tslot.raw");//' ../../LargeBarrelAnalysis/main.cpp
sed -i 's/manager.useTask("TimeCalibLoader", "tslot.raw", "tslot.calib");//' ../../LargeBarrelAnalysis/main.cpp

There is no TimeCalibLoader UserTask and TimeWindowCreator have tslot.calib output not tslot.raw.

How should I fix it?

wkrzemien:
I think @alekgajos should anwser it.

alekgajos:
The tests should be fixed after rebasing this PR on top of current develop. 

This was the matter of a deprecated setup for the integration tests in Travis and Jenkins config files. Since the LBA example refactor, the last test was silently failing as it was set to remove one of the modules from the analysis chain, but the module no longer existed in the chain after the refactor.

The failure was, however, not revealed due to the lack of error handling until this PR.

The config files for Travis and Jenkins should now be fixed in develop and should perform the test correctly.

wkrzemien:
This was accepted as #97

PR #99, opened 2018-08-07
 corrected streaming PR from KamilD #2

nenprio:


CosmicAnalysis/CMakeLists.txt
```diff
@@ -0,0 +1,110 @@
+# J-PET tool project using J-PET framework
+#
+# Author: Adam Strzelecki <adam.strzelecki@uj.edu.pl>
+#
+# Description:
+#   Builds project depending on J-PET framework.
+
+cmake_minimum_required(VERSION 2.6)
+
+######################################################################
```
alekgajos: These comments belong to the template but should be removed in examples derived from the template.

CosmicAnalysis/EventCategorizerCosmic.cpp
```diff
@@ -0,0 +1,127 @@
...
+  vector<JPetEvent> events;
+  if (auto timeWindow = dynamic_cast<const JPetTimeWindow* const>(fEvent)) 
+  {
+    uint n = timeWindow->getNumberOfEvents();
+    for (uint i = 0; i < n; ++i) {
+
+	const auto& event = dynamic_cast<const JPetEvent&>(timeWindow->operator[](i));
+	if (event.getHits().size() >= 1) 
+	{
+		vector<JPetHit> hits = event.getHits();
```
alekgajos: There are tabs and wrong depths of indentation, it would be good to run astyle on the files.

Imaging/CMakeLists.txt
```diff
@@ -0,0 +1,110 @@
+# J-PET tool project using J-PET framework
+#
+# Author: Adam Strzelecki <adam.strzelecki@uj.edu.pl>
+#
+# Description:
+#   Builds project depending on J-PET framework.
+
+cmake_minimum_required(VERSION 2.6)
+
+######################################################################
```
alekgajos: Comments left from the template (same remark as above)

PhysicAnalysis/CMakeLists.txt
```diff
@@ -0,0 +1,110 @@
+# J-PET tool project using J-PET framework
+#
+# Author: Adam Strzelecki <adam.strzelecki@uj.edu.pl>
+#
+# Description:
+#   Builds project depending on J-PET framework.
+
+cmake_minimum_required(VERSION 2.6)
+
+######################################################################
```
alekgajos: same remark as above

LargeBarrelAnalysis/EventCategorizerTools.cpp
```diff
@@ -246,3 +246,104 @@ double EventCategorizerTools::calculateTOF(const JPetHit& firstHit, const JPetHi
...
+        double transformedX = relativeAngles.at(1)+relativeAngles.at(0);
+        double transformedY = relativeAngles.at(1)-relativeAngles.at(0);
+	double timeDiff = fabs( thirdHit.getTime()/1000. - firstHit.getTime()/1000. );
+	double distanceFromCenter = calcDistanceOfSurfaceAndCenter(firstHit, secondHit, thirdHit);
+	if(saveHistos){
+	  stats.getHisto2D("DecayInto3_Angles")->Fill(transformedX, transformedY);
+	  stats.getHisto1D("DecayInto3_Distance")->Fill(distanceFromCenter);
+	  stats.getHisto1D("DecayInto3_TimeDiff")->Fill(timeDiff);
+	}
+	if( transformedX > d3SlotThetaMin && timeDiff < d3TimeDiff && distanceFromCenter < d3DistanceFromCenter )
```
alekgajos: Is the logic ok for sure? If this condition is not satisfied for any combination of 3 hits and the end of the triple loop is reached, the function would still return true?

CosmicAnalysis/EventCategorizerCosmic.cpp
```diff
@@ -0,0 +1,127 @@
...
+EventCategorizerCosmic::EventCategorizerCosmic(const char* name): JPetUserTask(name) {}
+
+bool EventCategorizerCosmic::init()
+{
+
+  INFO("Event categorization started.");
+  
+  fOutputEvents = new JPetTimeWindow("JPetEvent");
+  
+  if ( isOptionSet(fParams.getOptions(), kMinCosmicTOTParamKey) )
```
alekgajos: While not required, it is a good practice to add a warning in case an option was not set to make the user aware that a default value was used.
See LargeBarrelAnalysis for examples of this practice.

Imaging/EventCategorizerImaging.cpp
```diff
@@ -0,0 +1,246 @@
...
+EventCategorizerImaging::EventCategorizerImaging(const char* name): JPetUserTask(name) {}
+
+bool EventCategorizerImaging::init()
+{
+
+  INFO("Event categorization started.");
+  
+  fOutputEvents = new JPetTimeWindow("JPetEvent");
+    
+  if ( isOptionSet(fParams.getOptions(), kMinAnnihilationParamKey) )
```
alekgajos: Please consider issuing warnings in case the parameters are not set and defaults are used.

PhysicAnalysis/EventCategorizerPhysics.cpp
```diff
@@ -0,0 +1,314 @@
...
+EventCategorizerPhysics::EventCategorizerPhysics(const char* name): JPetUserTask(name) {}
+
+bool EventCategorizerPhysics::init()
+{
+
+  INFO("Event categorization started.");
+  
+  fOutputEvents = new JPetTimeWindow("JPetEvent");
+  
+  if ( isOptionSet(fParams.getOptions(), kMinAnnihilationParamKey) )
```
alekgajos: As above, please consider issuing warnings in case the parameters are not set and defaults are used.

wkrzemien:
I agree with all the remarks. Please, correct  it ASAP, so we can include it in next version.

CosmicAnalysis/EventCategorizerCosmic.cpp
```diff
@@ -0,0 +1,127 @@
...
+bool EventCategorizerCosmic::exec()
+{
+  vector<JPetEvent> events;
+  if (auto timeWindow = dynamic_cast<const JPetTimeWindow* const>(fEvent)) 
+  {
+    uint n = timeWindow->getNumberOfEvents();
+    for (uint i = 0; i < n; ++i) {
+
+	const auto& event = dynamic_cast<const JPetEvent&>(timeWindow->operator[](i));
+	if (event.getHits().size() >= 1)
```
kkacprzak: This if condition is redundant - each event has at least one hit.

PhysicAnalysis/EventCategorizerPhysics.cpp
```diff
@@ -0,0 +1,314 @@
...
+	physicEvent.addEventType(JPetEventType::kPrompt);
+      if( annihilationHits.getHits().size() > 0 )
+	{
+	  getStatistics().getHisto1D("Positronium_Lifetime")->Fill( annihilationHits.getHits().at(0).getTime()/1000. - deexcitationHits.getHits().at(0).getTime()/1000. );
+	}
+    }
+  if( EventCategorizerTools::checkFor2Gamma(annihilationHits, getStatistics(), fSaveControlHistos, fBackToBackAngleWindow, fMaxTimeDiff, fMaxDistOfDecayPlaneFromCenter) )
+    {
+      if(physicEvent.isOnlyTypeOf(JPetEventType::kUnknown))
+	physicEvent.setEventType(JPetEventType::k2Gamma);
```
kkacprzak: Same remark with adding types as in the imaging stream.

PR #102, opened 2018-08-19
Streaming improvements

kkacprzak:
- renaming of two methods used for streaming
- correcting names and definitions of the histograms
- removing unused argument in one of the streaming methods
- unification of the units - all  methods return in [ps], only in some histograms are filled with [ns]
- adding description of the parameters for streaming
- adding tests for streaming methods

### DIFF
--- a/PhysicAnalysis/EventCategorizerPhysics.cpp
+++ b/PhysicAnalysis/EventCategorizerPhysics.cpp
@@ -79,7 +79,7 @@ bool EventCategorizerPhysics::init()
 
   if (fSaveControlHistos) {
     getStatistics().createHistogram(
-      new TH1F("AllHitTOT", "TOT of all Hits in physics stream", 200, -0.5, 99.5)
+      new TH1F("AllHitTOT", "TOT of all Hits in physics stream", 200, 0.0, 100.0)
     );
     getStatistics().getHisto1D("AllHitTOT")->SetXTitle("TOT [ns]");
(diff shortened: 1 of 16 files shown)


PhysicAnalysis/EventCategorizerPhysics.cpp
```diff
@@ -27,208 +27,160 @@ EventCategorizerPhysics::EventCategorizerPhysics(const char* name): JPetUserTask
...
-    getStatistics().createHistogram(
-      new TH1F("Annih_TOF", "Annihilation pairs Time of Flight", 200, -3000.0, 3000.0));
-    getStatistics().getHisto1D("Annih_TOF")->GetXaxis()->SetTitle("Time of Flight [ps]");
-    getStatistics().getHisto1D("Annih_TOF")->GetYaxis()->SetTitle("Number of Annihilation Pairs");
-
     getStatistics().createHistogram(
-      new TH2F("AnnihPoint_XY", "XY position of annihilation point", 121, -60.5, 60.5, 121, -60.5, 60.5));
-    getStatistics().getHisto2D("AnnihPoint_XY")->GetXaxis()->SetTitle("X position [cm]");
-    getStatistics().getHisto2D("AnnihPoint_XY")->GetYaxis()->SetTitle("Y position [cm]");
+      new TH1F("AllHitTOT", "TOT of all Hits in physics stream", 200, -0.5, 99.5)
```
alekgajos: Same comment, zero between bins so the shift to negative values makes little sense.

PR #107, opened 2018-10-18
Changes for LLT case

kkacprzak:
Expansion includes:
- adding TimeWindowCreatorTools class with... tools for time window creation. 
- methods were extracted to tools class, including modifications of the Unpacker
- application of flagging Signal Channels with RecoFlags - repeated Leadings are Corrupted
- updating Signal Finder, Signal Transformer, Hit Finder and Event Finder with parameter reading for use/not use corrupted lower class structures
- few new control histograms
- more comments and descriptions


kkacprzak:
How to make it build with develop branch of Framework?

LargeBarrelAnalysis/SignalFinderTools.cpp
```diff
@@ -10,189 +10,178 @@
...
  */
-vector<JPetRawSignal> SignalFinderTools::buildRawSignals(
-  const vector<JPetSigCh>& sigChFromSamePM,
-  unsigned int numOfThresholds,
-  JPetStatistics& stats,
-  double sigChEdgeMaxTime,
-  double sigChLeadTrailMaxTime,
-  bool saveHistos)
-{
+ vector<JPetRawSignal> SignalFinderTools::buildRawSignals(
```
alekgajos: Is this method adapted to any order of threshold values?

LargeBarrelAnalysis/EventFinder.cpp
```diff
@@ -22,89 +22,152 @@ using namespace std;
...
+        event.setRecoFlag(JPetEvent::Corrupted);
+        getStatistics().getHisto1D("good_vs_bad_events")->Fill(2);
+      } else {
+        event.setRecoFlag(JPetEvent::Good);
+        getStatistics().getHisto1D("good_vs_bad_events")->Fill(1);
+      }
       eventVec.push_back(event);
-      if(fSaveControlHistos)
+      if(fSaveControlHistos) {
         getStatistics().getHisto1D("hits_per_event")->Fill(event.getHits().size());
```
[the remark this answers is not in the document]
kkacprzak: IIUC (if I understand correctly) - so histogram could be filled with all information about multiplicity, event if i.e. I put condition that I want only 3 hit events saved, 1-hit and 2-hit bins will be non-empty. Wouldn't that be misleading? Someone sets flag for 3-hit and then sees multiplicity filled with others. 
Solution would be to make two histograms - multiplicity_all and multiplicity_selected, or something like this. I can do this.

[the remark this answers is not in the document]
kkacprzak: same as above

LargeBarrelAnalysis/HitFinderTools.cpp
```diff
@@ -23,104 +23,124 @@ using namespace std;
...
+    return emptyVec;
   }
+  // Init return vector
+  int slotNumber = paramBank.getBarrelSlotsSize();
+  vector<JPetPhysSignal> signalVec;
+  vector<vector<JPetPhysSignal>> slotSignalVec(slotNumber, signalVec);
+  // Map Signals according to Slot they belong to
+  const unsigned int nSigals = timeWindow->getNumberOfEvents();
+  for (unsigned int i = 0; i < nSigals; i++) {
+    auto sigal = dynamic_cast<const JPetPhysSignal&>(timeWindow->operator[](i));
```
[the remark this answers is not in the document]
kkacprzak: aa, typo, sure

LargeBarrelAnalysis/HitFinderTools.cpp
```diff
@@ -172,34 +206,23 @@ JPetHit HitFinderTools::createDummyRefDetHit(const JPetPhysSignal& signalB)
...
 /**
  * Helper method for getting TOMB channel for velocity map
  */
 int HitFinderTools::getProperChannel(const JPetPhysSignal& signal)
 {
-  JPetSigCh someSigCh = signal.getRecoSignal().getRawSignal()
-                        .getPoints(JPetSigCh::Leading, JPetRawSignal::ByThrNum)[0];
+  auto someSigCh = signal.getRecoSignal().getRawSignal()
+    .getPoints(JPetSigCh::Leading, JPetRawSignal::ByThrValue)[0];
+  // INFO(someSigCh.getTOMBChannel().GetObject());
```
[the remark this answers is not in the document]
kkacprzak: removed

LargeBarrelAnalysis/TimeWindowCreator.h
```diff
@@ -50,23 +48,26 @@ class TimeWindowCreator: public JPetUserTask
...
-	JPetSigCh generateSigCh(const JPetTOMBChannel& channel, JPetSigCh::EdgeType edge) const;
-	const std::string kMainStripKey = "TimeWindowCreator_MainStrip_int";
-	const std::string kMaxTimeParamKey = "TimeWindowCreator_MaxTime_float";
-	const std::string kMinTimeParamKey = "TimeWindowCreator_MinTime_float";
+	void saveSigChs(const std::vector<JPetSigCh>& sigChVec);
+	bool filter(JPetTOMBChannel& tombChannel) const;
+	void initialiseHistograms();
 	const std::string kTimeCalibFileParamKey = "TimeCalibLoader_ConfigFile_std::string";
 	const std::string kThresholdFileParamKey = "ThresholdLoader_ConfigFile_std::string";
 	const std::string kSaveControlHistosParamKey = "Save_Cotrol_Histograms_bool";
```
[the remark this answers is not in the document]
kkacprzak: It seems that this typo is there from the beginning :) And it is also in the `userParams.json` file on sphinx. Soooo... we should change this file and maybe make new directory on the server with files, something like ExamplesV7

LargeBarrelAnalysis/SignalTransformer.h
```diff
@@ -23,26 +23,31 @@
...
 	virtual bool init() override;
 	virtual bool exec() override;
 	virtual bool terminate() override;
 
 protected:
+	void initialiseHistograms();
 	JPetRecoSignal createRecoSignal(const JPetRawSignal& rawSignal);
 	JPetPhysSignal createPhysSignal(const JPetRecoSignal& signals);
+	const std::string kUseCorruptedSignalsParamKey = "SignalTransformer_UseCorruptedSignals_bool";
+	const std::string kSaveControlHistosParamKey = "Save_Cotrol_Histograms_bool";
```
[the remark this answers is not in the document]
kkacprzak: same as above

LargeBarrelAnalysis/EventCategorizer.h
```diff
@@ -40,14 +40,14 @@ class JPetWriter;
...
 	virtual bool exec() override;
 	virtual bool terminate() override;
 
 protected:
+	const std::string kBack2BackSlotThetaDiffParamKey = "Back2Back_Categorizer_SlotThetaDiff_float";
 	const std::string kScatterTOFTimeDiffParamKey = "Scatter_Categorizer_TOF_TimeDiff_float";
-  const std::string kBack2BackSlotThetaDiffParamKey = "Back2Back_Categorizer_SlotThetaDiff_float";
 	const std::string kDeexTOTCutMinParamKey = "Deex_Categorizer_TOT_Cut_Min_float";
 	const std::string kDeexTOTCutMaxParamKey = "Deex_Categorizer_TOT_Cut_Max_float";
 	const std::string kSaveControlHistosParamKey = "Save_Cotrol_Histograms_bool";
```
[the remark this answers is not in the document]
kkacprzak: corrected

LargeBarrelAnalysis/EventFinder.h
```diff
@@ -35,24 +35,28 @@ class JPetWriter;
...
   virtual bool terminate() override;
 
 protected:
-  void saveEvents(const std::vector<JPetEvent>& event);
   std::vector<JPetEvent> buildEvents(const JPetTimeWindow & hits);
+  void saveEvents(const std::vector<JPetEvent>& event);
+  void initialiseHistograms();
+  const std::string kUseCorruptedHitsParamKey = "EventFinder_UseCorruptedHits_bool";
+  const std::string kEventMinMultiplicity = "EventFinder_MinEventMultiplicity_int";
   const std::string kSaveControlHistosParamKey = "Save_Cotrol_Histograms_bool";
```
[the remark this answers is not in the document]
kkacprzak: same...

LargeBarrelAnalysis/HitFinder.h
```diff
@@ -39,21 +39,24 @@ class HitFinder: public JPetUserTask {
...
 
 protected:
   void saveHits(const std::vector<JPetHit>& hits);
+  void initialiseHistograms();
   std::map<unsigned int, std::vector<double>> fVelocities;
-  const std::string kSaveControlHistosParamKey = "Save_Cotrol_Histograms_bool";
+  const std::string kUseCorruptedSignalsParamKey = "HitFinder_UseCorruptedSignals_bool";
   const std::string kVelocityFileParamKey = "HitFinder_VelocityFile_std::string";
-  const std::string kABTimeDiffParamKey = "HitFinder_ABTimeDiff_float";
+  const std::string kSaveControlHistosParamKey = "Save_Cotrol_Histograms_bool";
```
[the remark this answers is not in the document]
kkacprzak: Interesting that nobody spotted that before

LargeBarrelAnalysis/SignalFinder.h
```diff
@@ -46,14 +44,16 @@ class SignalFinder: public JPetUserTask
 
 protected:
   void saveRawSignals(const std::vector<JPetRawSignal>& sigChVec);
-  const std::string kEdgeMaxTimeParamKey = "SignalFinder_EdgeMaxTime_float";
+  const std::string kUseCorruptedSigChParamKey = "SignalFinder_UseCorruptedSigCh_bool";
   const std::string kLeadTrailMaxTimeParamKey = "SignalFinder_LeadTrailMaxTime_float";
   const std::string kSaveControlHistosParamKey = "Save_Cotrol_Histograms_bool";
```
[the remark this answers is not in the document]
kkacprzak: suprising

LargeBarrelAnalysis/SignalFinderTools.cpp
kkacprzak: see general comment

kkacprzak:
ad. 1. In this PR I added only changes for detection and handling of LLT problem. Order of thresholds is a different issue, which I would like to solve otherwise. I my opinion we should rely on configuration files for proper description of the setup, and this is the case. Changing the thresholds is a change of setup, so new json file should be produced and applied. Inside the json file indices of threshold should follow increasing values of set voltage. I am not a supporter of changing the universal code for single cases, right now the signal finding methods (and others) are general enough so just proper configuration is required. 

ad. 2 and 3. I think this is a case for separate PR, which would include mentioned optimization ideas. So 1) substitute (?) maps in UniversalFileLoader and methods for finding calibration values 2) proper distribution of data to PM or slots in the modules 3) smarter usage of TOMB channels?

kkacprzak:
Also I would like to get some feedback after tests and about the logic of setting the reco flags.

LargeBarrelAnalysis/EventFinder.cpp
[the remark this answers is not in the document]
alekgajos: Yes, this is what I meant. Otherwise the histogram is the number of hits per event saved to the output but not the situation in all the data processed by EventFinder. If for some reason, there were no 1-hit events (e.g. if the user accidentally set the event time window to 1e7 ps), then the user would not notice that;-)

You are right that the best way would be to create two histograms, for all events and for the passed events only.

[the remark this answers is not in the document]
alekgajos: Oops, sorry for the duplicated comment.

alekgajos:
Ad. 1. Thank for the clarification. I was confused because in HitFinder and SignalTransformer you are already using sorting of SigChs by threshold value.
I don't fully like the idea of doing it at the configuration level only- as you said, this requires to change the local threshold number corresponding to a particular DAQ channel in the configuration file, but this would lead to having config files not reflecting the true state of the DAQ system. The local channel number is not just a convention - it corresponds to a particluar and fixed output channel of the front-end board, having e.g. specific time delays. Mixing those numbers can lead to possible confusion.
@wictus has already prepared the JSON files where the local threshold numbers are unchanged by the actual set threshold values are adaoted to match the settings of particular sub-runs:
http://koza.if.uj.edu.pl/petwiki/index.php/Default_settings_and_parameters_used_in_the_analyses
I think this is the right way to go.
Anyway, SignalFinder can be adapted to this solution with minimal modification only - I think is is enough to create a permutation of the channel numbers based on the increasing threshold values and then in lines 88-90 do something like (if **P** is the permutation):

```cpp
    thrLeadingSigCh.at( P(sigCh.getThresholdNumber()) - 1 ).push_back(sigCh);
 ...
    thrTrailingSigCh.at( P(sigCh.getThresholdNumber()) - 1 ).push_back(sigCh);
```
We can do it in a separate PR.

alekgajos:
Ad. 2. Ok, this indeed is a minor optimization and we can do it separately.

Ad. 3. This, however, is a more serious problem. Changing the maps to vectors in the current way we would break the universality of the code we had until now and I think we should not do it. It is better not to change the old solution in a partial way, so I propose to leave the use of maps for now.
Then in a separate PR we can carefully optimize such cases.

kkacprzak:
Yes Alek, you are right, it is not worth it to make partial changes. Lets make some optimization PR later. Now I will commit these minor changes and go back to maps.

LargeBarrelAnalysis/HitFinderTools.cpp
```diff
@@ -23,104 +23,124 @@ using namespace std;
...
+/**
+ * Loop over all Scins invoking matching procedure
+ */
+vector<JPetHit> HitFinderTools::matchAllSignals(
+  const vector<vector<JPetPhysSignal>>& allSignals,
+  const map<unsigned int, vector<double>>& velocitiesMap,
+  double timeDiffAB, int refDetScinId, JPetStatistics& stats, bool saveHistos
+) {
+  vector<JPetHit> allHits;
+  for (auto slotSigals : allSignals) {
```
wkrzemien: for (auto& slotSignals: allSignals) {
...

kkacprzak:
I added suggested corrections and more: 
- returned to mapping PM/Slots with maps, not simple vectors <- main (un)change
- small typos, but not with "control histograms" -> other small PR upcoming
- slight modification of LLT histograms
- two multiplicity histograms in Event Finder before and after cut
- adding new parameters descriptions 

I have tested usage of reco flags and i.e. in files from Run 5 less than 1% of SigChs are corrupted, and it propagates also to less than 1% of signals, hits and events. I set default values for UseCorrupted flags to false, and this indeed cuts all corrupted objects.

kkacprzak:
P.S. I will commit today one more time - small corrections to tests

wkrzemien:
Hi, Do you know why the tests are not currently passing?

LargeBarrelAnalysis/EventCategorizer.cpp
```diff
@@ -22,37 +22,61 @@
 using namespace jpet_options_tools;
 using namespace std;
 
+/**
```
wkrzemien: Those kind of comments make not much sense IMHO.
I would vote for removing Constructor, destructor comments

LargeBarrelAnalysis/EventCategorizer.cpp
```diff
@@ -22,37 +22,61 @@
...
+ */
 EventCategorizer::EventCategorizer(const char* name): JPetUserTask(name) {}
 
+/**
+ * Destructor
+ */
+EventCategorizer::~EventCategorizer() {}
+
+/**
+ * Init Event Categorizer
```
wkrzemien: I think the function name is self-explanatory. The comment is not needed
I would recommend to have a look here:
http://apdevblog.com/comments-in-code/
Or Robert C. Martin "Clean code" book

LargeBarrelAnalysis/EventCategorizer.cpp
```diff
@@ -102,9 +138,12 @@ bool EventCategorizer::terminate()
 
 void EventCategorizer::saveEvents(const vector<JPetEvent>& events)
 {
-  for (const auto& event : events) fOutputEvents->add<JPetEvent>(event);
+  for (const auto& event : events) { fOutputEvents->add<JPetEvent>(event); }
 }
 
+/**
+ * Init histograms
```
wkrzemien: i would remove this comment :-)

LargeBarrelAnalysis/EventCategorizerTools.h
```diff
@@ -13,12 +13,13 @@
...
-#ifndef _EVENTCATEGORIZERTOOLS_H_
-#define _EVENTCATEGORIZERTOOLS_H_
+#ifndef EVENTCATEGORIZERTOOLS_H
+#define EVENTCATEGORIZERTOOLS_H
 
 #include <JPetStatistics/JPetStatistics.h>
 #include <JPetEvent/JPetEvent.h>
 #include <JPetHit/JPetHit.h>
+
 #define kLightVelocity_cm_ns 29.9792458
```
wkrzemien: Since we are polishing everything
Could you  change it to class members:
public static const double kLight ...
there is no reason to use #define

LargeBarrelAnalysis/EventFinder.cpp
```diff
@@ -22,89 +22,172 @@ using namespace std;
 
 using namespace jpet_options_tools;
 
+/**
+ * Constructor
```
wkrzemien: the same as above

LargeBarrelAnalysis/EventFinder.cpp
```diff
@@ -22,89 +22,172 @@ using namespace std;
...
-  if (fSaveControlHistos){
-    getStatistics().createHistogram(
-      new TH1F("hits_per_event", "Number of Hits in Event", 20, 0.5, 20.5));
-    getStatistics().getHisto1D("hits_per_event")
-      ->GetXaxis()->SetTitle("Hits in Event");
-    getStatistics().getHisto1D("hits_per_event")
-      ->GetYaxis()->SetTitle("Number of Hits");
   }
 
+  // Initialize histograms
```
wkrzemien: to remove

LargeBarrelAnalysis/EventFinder.cpp
```diff
@@ -22,89 +22,172 @@ using namespace std;
...
+ * Terminate Event Finder
+ */
 bool EventFinder::terminate()
 {
   INFO("Event fiding ended.");
   return true;
 }
 
+/**
+ * Saving method
```
wkrzemien: to remove

LargeBarrelAnalysis/EventFinder.cpp
```diff
@@ -22,89 +22,172 @@ using namespace std;
...
-
+    if(hit.getRecoFlag() == JPetHit::Good) {
+      event.setRecoFlag(JPetEvent::Good);
+    } else if(hit.getRecoFlag() == JPetHit::Corrupted){
+      event.setRecoFlag(JPetEvent::Corrupted);
+    }
     unsigned int nextCount = 1;
-    while(count+nextCount<nHits){
-      JPetHit nextHit = dynamic_cast<const JPetHit&>(timeWindow.operator[](count+nextCount));
+    while(count+nextCount < nHits){
```
wkrzemien: Here the comment would be useful, cause it is not so simple, to understand what is going on.

LargeBarrelAnalysis/EventFinder.cpp
```diff
@@ -22,89 +22,172 @@ using namespace std;
...
+      if(fSaveControlHistos) {
+        getStatistics().getHisto1D("hits_per_event_selected")->Fill(event.getHits().size());
+      }
     }
   }
   return eventVec;
 }
+
+/**
+ * Init histograms
```
wkrzemien: to remove

LargeBarrelAnalysis/EventFinder.h
wkrzemien: Do we leave this typo for the future?

LargeBarrelAnalysis/HitFinder.cpp
```diff
@@ -28,33 +28,58 @@ using namespace std;
 
 using namespace jpet_options_tools;
 
-HitFinder::HitFinder(const char* name) : JPetUserTask(name) { }
+/**
+ * Constructor
```
wkrzemien: to remove

LargeBarrelAnalysis/HitFinder.cpp
```diff
@@ -28,33 +28,58 @@ using namespace std;
...
+ */
+HitFinder::HitFinder(const char* name) : JPetUserTask(name) {}
+
+/**
+ * Destructor
```
wkrzemien: remove

LargeBarrelAnalysis/HitFinder.cpp
```diff
@@ -28,33 +28,58 @@ using namespace std;
...
+ */
+HitFinder::~HitFinder() {}
 
+/**
+ * Init Hit Finder
```
wkrzemien: remove

LargeBarrelAnalysis/HitFinderTools.cpp
```diff
@@ -52,75 +63,88 @@ map<int, vector<JPetPhysSignal>> HitFinderTools::getSignalsSlotMap(
...
+  vector<JPetPhysSignal> remainSignals;
+  sortByTime(slotSignals);
+  while (slotSignals.size() > 0) {
+    auto physSig = slotSignals.at(0);
+    if(slotSignals.size() == 1){
+      remainSignals.push_back(physSig);
+      break;
+    }
+    for (unsigned int j = 1; j < slotSignals.size(); j++) {
+      if (slotSignals.at(j).getTime() - physSig.getTime() < timeDiffAB) {
```
wkrzemien: this difference should not be abs value?

LargeBarrelAnalysis/HitFinderTools.cpp
```diff
@@ -173,8 +217,18 @@ JPetHit HitFinderTools::createDummyRefDetHit(const JPetPhysSignal& signalB)
 }
 
 /**
- * Helper method for checking if theta is in radians
+ * Helper method for getting TOMB channel for velocity map
```
wkrzemien: I dont see velocity map used in this function?
Maybe just for getting TOMB channel. is enough

LargeBarrelAnalysis/HitFinderTools.h
```diff
@@ -18,6 +18,7 @@
 
 #include <JPetStatistics/JPetStatistics.h>
 #include <JPetTimeWindow/JPetTimeWindow.h>
+#include <JPetParamBank/JPetParamBank.h>
```
wkrzemien: why do we need it here?

LargeBarrelAnalysis/SignalTransformer.cpp
```diff
@@ -16,31 +16,112 @@
 #include "JPetWriter/JPetWriter.h"
 #include "SignalTransformer.h"
 
+using namespace jpet_options_tools;
+
+/**
+ * Constructor
```
wkrzemien: to remove

LargeBarrelAnalysis/SignalTransformer.cpp
```diff
@@ -16,31 +16,112 @@
...
+ */
 SignalTransformer::SignalTransformer(const char* name): JPetUserTask(name) {}
 
+/**
+ * Destructor
```
wkrzemien: to remove

LargeBarrelAnalysis/SignalTransformer.cpp
```diff
@@ -16,31 +16,112 @@
...
+ */
+SignalTransformer::~SignalTransformer() {}
+
+/**
+ * Init Signal Transformer
```
wkrzemien: the same, in previous you used intialize here init, you see how it starts to diverge.

LargeBarrelAnalysis/SignalTransformer.cpp
```diff
@@ -16,31 +16,112 @@
...
+    }
+  } else {
+    WARNING("Signal Transformer is not using Corrupted Signals (default option)");
+  }
+  // Getting bool for saving histograms
+  if (isOptionSet(fParams.getOptions(), kSaveControlHistosParamKey)) {
+    fSaveControlHistos = getOptionAsBool(fParams.getOptions(), kSaveControlHistosParamKey);
+  }
+
+  // Control histograms
```
wkrzemien: to remove

LargeBarrelAnalysis/SignalTransformer.cpp
```diff
@@ -58,24 +139,109 @@ JPetRecoSignal SignalTransformer::createRecoSignal(const JPetRawSignal& rawSigna
...
-  JPetSigCh::Leading, JPetRawSignal::ByThrNum);
+    JPetSigCh::Leading, JPetRawSignal::ByThrValue
+  );
   physSignal.setTime(leadingSigChVec.at(0).getValue());
-  physSignal.setQualityOfTime(0.0);
   return physSignal;
 }
+
+/**
+ * Init histograms
```
wkrzemien: to remove

LargeBarrelAnalysis/TimeWindowCreator.cpp
```diff
@@ -16,14 +16,27 @@
...
+#include "TimeWindowCreatorTools.h"
 #include <JPetWriter/JPetWriter.h>
 #include "UniversalFileLoader.h"
 #include "TimeWindowCreator.h"
 
 using namespace jpet_options_tools;
+using namespace std;
 
+/**
+ * Constructor
```
wkrzemien: to remove

LargeBarrelAnalysis/TimeWindowCreator.cpp
```diff
@@ -16,14 +16,27 @@
...
+ */
 TimeWindowCreator::TimeWindowCreator(const char* name): JPetUserTask(name) {}
 
+/**
+ * Destructor
```
wkrzemien: the same

LargeBarrelAnalysis/TimeWindowCreator.cpp
```diff
@@ -16,14 +16,27 @@
...
+ */
+TimeWindowCreator::~TimeWindowCreator() {}
+
+/**
+ * Init Time Window Creator
```
wkrzemien: the same

LargeBarrelAnalysis/TimeWindowCreator.cpp
```diff
@@ -102,123 +124,165 @@ bool TimeWindowCreator::init()
...
-    getStatistics().createHistogram(
-      new TH1F("sig_ch_per_time_slot",
-               "Signal Channels Per Time Slot",
-               250, -0.5, 999.5));
-    getStatistics().getHisto1D("sig_ch_per_time_slot")
-    ->GetXaxis()->SetTitle("Signal Channels in Time Slot");
-    getStatistics().getHisto1D("sig_ch_per_time_slot")
-    ->GetYaxis()->SetTitle("Number of Time Slots");
-  }
+  // Control histograms
```
wkrzemien: remove

LargeBarrelAnalysis/TimeWindowCreator.cpp
```diff
@@ -102,123 +124,165 @@ bool TimeWindowCreator::init()
...
+  if (fSaveControlHistos) { initialiseHistograms(); }
   return true;
 }
 
-TimeWindowCreator::~TimeWindowCreator() {}
-
+/**
+ * Execute Time Window Creator
```
wkrzemien: this is also not necessary

LargeBarrelAnalysis/TimeWindowCreator.cpp
```diff
@@ -102,123 +124,165 @@ bool TimeWindowCreator::init()
...
-  return sigCh;
+  for (auto & sigCh : sigChVec) { fOutputEvents->add<JPetSigCh>(sigCh); }
 }
 
 /**
  * Reference Detector
  * Returns true if signal from the channel given as argument should be passed
  */
-bool TimeWindowCreator::filter(const JPetTOMBChannel& channel) const
+bool TimeWindowCreator::filter(JPetTOMBChannel& tombChannel) const
```
wkrzemien: the name of this method does not say anything about what it does.
isAllowedChannel() or something similar would be much better

LargeBarrelAnalysis/TimeWindowCreator.cpp
```diff
@@ -102,123 +124,165 @@ bool TimeWindowCreator::init()
...
-  if (fAllowedChannels.find(channel.getChannel())
-      != fAllowedChannels.end()) return true;
+  if (fAllowedChannels.find(tombChannel.getChannel()) != fAllowedChannels.end()) {
+    return true;
+  }
   return false;
 }
+
+/**
+ * Init histograms
```
wkrzemien: to remove

LargeBarrelAnalysis/TimeWindowCreator.h
```diff
@@ -50,23 +48,26 @@ class TimeWindowCreator: public JPetUserTask
 	virtual bool terminate() override;
 
 protected:
-	bool filter(const JPetTOMBChannel& channel) const;
-	JPetSigCh generateSigCh(const JPetTOMBChannel& channel, JPetSigCh::EdgeType edge) const;
-	const std::string kMainStripKey = "TimeWindowCreator_MainStrip_int";
-	const std::string kMaxTimeParamKey = "TimeWindowCreator_MaxTime_float";
-	const std::string kMinTimeParamKey = "TimeWindowCreator_MinTime_float";
+	void saveSigChs(const std::vector<JPetSigCh>& sigChVec);
+	bool filter(JPetTOMBChannel& tombChannel) const;
```
wkrzemien: if possible change the name of this method as above

LargeBarrelAnalysis/TimeWindowCreatorTools.cpp
```diff
@@ -0,0 +1,156 @@
...
+) {
+  for(unsigned int i=0; i<inputSigChs.size(); i++) {
+    if(i == inputSigChs.size()-1) {
+      inputSigChs.at(i).setRecoFlag(JPetSigCh::Good);
+      if(saveHistos){ stats.getHisto1D("good_vs_bad_sigch")->Fill(1); }
+      break;
+    }
+    auto& sigCh1 = inputSigChs.at(i);
+    auto& sigCh2 = inputSigChs.at(i+1);
+    // Explicit check for repeteated edges
```
wkrzemien: repeated

LargeBarrelAnalysis/TimeWindowCreatorTools.cpp
```diff
@@ -0,0 +1,156 @@
...
+    if (saveHistos){
+      stats.getHisto1D(Form("pm_occupation_thr%d", tombChannel.getLocalChannelNumber()))
+        ->Fill(tombChannel.getPM().getID());
+    }
+  }
+  return allTDCSigChs;
+}
+
+/**
+ * Method for investigation of repetated edges - setting RecoFlag for each SigCh
```
wkrzemien: Here some description would be nice

LargeBarrelAnalysis/EventCategorizerTools.h
kkacprzak: done

LargeBarrelAnalysis/EventFinder.h
kkacprzak: done

LargeBarrelAnalysis/HitFinderTools.cpp
kkacprzak: here no. Signals are sorted in time, so there is no use in checking backwards, it was checked with previous signal in this vector

LargeBarrelAnalysis/HitFinderTools.h
kkacprzak: leftover, removed

LargeBarrelAnalysis/TimeWindowCreator.h
kkacprzak: done

LargeBarrelAnalysis/HitFinderTools.cpp
wkrzemien: ok. thx

LargeBarrelAnalysis/EventCategorizer.h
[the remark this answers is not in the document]
wkrzemien: I still see "SaveCotrol_Histogram_bool"

kkacprzak:
PR updated, please check if framework submodule now points to proper branch/commit.

LargeBarrelAnalysis/HitFinderTools.cpp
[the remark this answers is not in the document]
wkrzemien: This typo remained

PR #110, opened 2018-12-11
Universal output for velocity calib

wictus:
Make output from velocity calibration the same as for UniversalCalib (same as in TimeCalibration)


VelocityCalibration/DeltaTFinder.cpp
```diff
@@ -75,6 +75,11 @@ bool DeltaTFinder::init()
     pos += "_std::string";
     if (isOptionSet(fParams.getOptions(), pos)) {
       auto res = retrievePositionAndFileName(getOptionAsString(fParams.getOptions(), pos));
+      if ( file_path.find_last_of("/") ) 
+	file_path = file_path.substr( file_path.find_last_of("/")+1,  file_path.find(".") - file_path.find_last_of("/")-1 );
```
wkrzemien: please add braces or at least one tab

wkrzemien: I am not 100% sure what are you doing here, but check JPetCommonTools maybe there is a method that can be used here, Also, boost:filesystem has a set of methods to deal with pahts

VelocityCalibration/estimateVelocity.cpp
```diff
@@ -159,6 +172,27 @@ int findScintillatorIn( std::vector<scintData>& data, int ID)
 	return -1;
 }
 
+bool ifThresholdIsIn( scintData& data, char label)
+{
```
wkrzemien: consider something like
return std::any_of(data.cbegin(), data.cend(), [label]( const TYPE_OF_THRESHOLD& thr) {return label == threshold.thresholdLabel; });

VelocityCalibration/DeltaTFinder.cpp
wictus: I needed filename, without path and extensions. I used JPetCommonTools::extractFileNameFromFullPath as you suggested.

VelocityCalibration/estimateVelocity.cpp
```diff
@@ -87,37 +87,50 @@ void parserTest(std::vector<scintData>& data, const char thrLabel)
...
-					positionsForFit.push_back( position.positionValue );
-					timesForFit.push_back( position.deltaT );
-					errorsForFit.push_back( position.deltaTError);
-					std::cout << positionsForFit.size() << std::endl;
-					std::cout << timesForFit.size() << std::endl;
-				}
+			std::vector<double> positionsForFit;
+			std::vector<double> timesForFit;	
+			std::vector<double> errorsForFit;
+			for( auto position : threshold.positions )
```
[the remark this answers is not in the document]
wictus: Changed, I only go through those structures and extract what I need.

wkrzemien:
retest this please

wictus:
I did one small fix, now it works with data properly.
Thanks for a very nice improvement in estimateVelocity.cpp

grey277:
run velocity tests

wkrzemien:
Why those tests are not passing? @wictus cause of the new reference files needed?

grey277:
run velocity tests

grey277:
retest this please

wkrzemien:
If I understand correctly @kkacprzak  there is an issue with user params. Can you fix it?

wictus:
Userparams.json was removed from repo

PR #112, opened 2018-12-18
Smal fixes for GeanAnalysis example

daria137:
- add Parameters description
- removed log cleaning
- add info to README


daria137:
new pull request will be open - with less messy history and without changing framework core

PR #137, opened 2019-06-04
LBA Tests check and fix

kkacprzak:
- revision of test in LargeBarrelAnalysis after other PRs 
- small correction in LLT flagging, so it works with the tests
- refactor of one variable name, so it follows convention
- small correction and addition in histogram in HitFinder

### DIFF
--- a/LargeBarrelAnalysis/TimeWindowCreatorTools.cpp
+++ b/LargeBarrelAnalysis/TimeWindowCreatorTools.cpp
@@ -100,9 +100,11 @@ void TimeWindowCreatorTools::flagSigChs(
         stats.getHisto1D("good_vs_bad_sigch")->Fill(1, 2);
       }
     } else if (sigCh1.getType() == JPetSigCh::Trailing && sigCh2.getType() == JPetSigCh::Leading) {
-      sigCh1.setRecoFlag(JPetSigCh::Good);
-      if(saveHistos){
-        stats.getHisto1D("good_vs_bad_sigch")->Fill(1);
+      if(sigCh1.getRecoFlag() == JPetSigCh::Unknown){
+        sigCh1.setRecoFlag(JPetSigCh::Good);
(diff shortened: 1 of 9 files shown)


LargeBarrelAnalysis/TimeWindowCreatorTools.cpp
```diff
@@ -100,9 +100,11 @@ void TimeWindowCreatorTools::flagSigChs(
         stats.getHisto1D("good_vs_bad_sigch")->Fill(1, 2);
       }
     } else if (sigCh1.getType() == JPetSigCh::Trailing && sigCh2.getType() == JPetSigCh::Leading) {
-      sigCh1.setRecoFlag(JPetSigCh::Good);
-      if(saveHistos){
-        stats.getHisto1D("good_vs_bad_sigch")->Fill(1);
+      if(sigCh1.getRecoFlag() == JPetSigCh::Unknown){
+        sigCh1.setRecoFlag(JPetSigCh::Good);
```
wkrzemien: I dont know the details, but why you set RecoFlag as Good if it was Unknown?
Is it what is supposed to  be ?
Maybe add some comment here?

wkrzemien: @alekgajos please have a look

alekgajos: This is during flagging of SigCh-s (recognizing LLT sequences) so I understand that on that stage "Unknown" is equivalent to "not yet flagged in previous steps" and this is probably ok. However, I don't know why it was necessary to add it now.

kkacprzak: Yes, this is the case when Trailing was not marked previously - in this if statement is is followed by Leading, so we mark it Good, later in the code there is a similar check for TT case.

PR #147, opened 2019-11-10
V5fixes

alekgajos:
* Removed obsoleted parameter from MCGeantAnalysis/PARAMETERS.md and added information about smearing possibility
* Fixed a bug crashing LargeBarrelAnalysis due to a histogram referenced but not created in the module init (thus causing a segmentation fault).


alekgajos:
No problem, description updated.

PR #151, opened 2020-03-04
Expanding TOT calculations to apply thresholds

kdulski:
Expanding TOT calculations to apply thresholds from the config file.
calculateTOT(const JPetHit& hit) changed to 
calculateTOT(const JPetHit& hit, TOTCalculationType type)
where type stands for different way of applying thresholds. 

Added definition of TOTCalculationType as 
enum TOTCalculationType
  {
    kStandard,
    kAdjustedToThresholds,
    kAdjustedToThresholdsExtended
  }; 

which stands for:
 - kStandard : TOT as a simple sum of TOT over all thresholds
 - kAdjustedToThresholds : TOT calculated as a rectangulars, thresholds used as a height of a rectangular
 - kAdjustedToThresholdsExtended : TOT as above with addition of triangles calculated between thresholds

weight parameter introduced to obtain similiar values of TOT for different TOTCalculationsType.
For 
kStandard : weight = 1
kAdjustedToThresholds and kAdjustedToThresholdsExtended : 
weight = threshold(i) - threshold(i-1)/threshold(1) // threshold(0) = 0
threshold(1) - first threshold used as a reference, because every signal has it and it is easier to implement that to use maximal difference between thresholds.

### DIFF
--- a/LargeBarrelAnalysis/EventCategorizerTools.cpp
+++ b/LargeBarrelAnalysis/EventCategorizerTools.cpp
@@ -111,7 +111,7 @@ bool EventCategorizerTools::checkForPrompt(
   double deexTOTCutMin, double deexTOTCutMax)
 {
   for (unsigned i = 0; i < event.getHits().size(); i++) {
-    double tot = calculateTOT(event.getHits().at(i), TOTCalculationType::kStandard);
+    double tot = calculateTOT(event.getHits().at(i), TOTCalculationType::kSimplified);
     if (tot > deexTOTCutMin && tot < deexTOTCutMax) {
       if (saveHistos) {
         stats.getHisto1D("Deex_TOT_cut")->Fill(tot);
@@ -153,8 +153,8 @@ bool EventCategorizerTools::checkForScatter(
 
       if (fabs(scattTOF - timeDiff) < scatterTOFTimeDiff) {
         if (saveHistos) {
-          stats.getHisto2D("ScatterAngle_PrimaryTOT")->Fill(scattAngle, calculateTOT(primaryHit, TOTCalculationType::kStandard));
-          stats.getHisto2D("ScatterAngle_ScatterTOT")->Fill(scattAngle, calculateTOT(scatterHit, TOTCalculationType::kStandard));
+          stats.getHisto2D("ScatterAngle_PrimaryTOT")->Fill(scattAngle, calculateTOT(primaryHit, TOTCalculationType::kSimplified));
+          stats.getHisto2D("ScatterAngle_ScatterTOT")->Fill(scattAngle, calculateTOT(scatterHit, TOTCalculationType::kSimplified));
         }
         return true;
       }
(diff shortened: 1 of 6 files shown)


kdulski:
correction
weight = (threshold(i) - threshold(i-1))/threshold(1)

LargeBarrelAnalysis/EventCategorizerTools.cpp
```diff
@@ -179,15 +180,41 @@ double EventCategorizerTools::calculateTOT(const JPetHit& hit)
                    .getPoints(JPetSigCh::Trailing, JPetRawSignal::ByThrNum);
   auto sigBTrail = hit.getSignalB().getRecoSignal().getRawSignal()
                    .getPoints(JPetSigCh::Trailing, JPetRawSignal::ByThrNum);
-
+                   
   if (sigALead.size() > 0 && sigATrail.size() > 0) {
     for (unsigned i = 0; i < sigALead.size() && i < sigATrail.size(); i++) {
```
kkacprzak: ok, this part could be more clear, so I would do:
```c++
// thr1 by default
tot += (sigATrail.at(0).getValue() - sigALead.at(0).getValue());
// now the loop starting from i=1
for (unsigned i = 1; i < sigALead.size() && i < sigATrail.size(); i++) {
 // use switch 
  switch(type) {
    case TOTCalculationType::kStandard:
      ... 
    case TOTCalculationType::kAdjustedToThresholds:
      ...
    case TOTCalculationType::kAdjustedToThresholdsExtended:
      ...
  }
}
```

LargeBarrelAnalysis/EventCategorizerTools.cpp
```diff
@@ -179,15 +180,41 @@ double EventCategorizerTools::calculateTOT(const JPetHit& hit)
...
+        else
+          weight = 1.;
+      }
+      else 
+        weight = 1.;
+      tot += weight*(sigATrail.at(i).getValue() - sigALead.at(i).getValue());
     }
   }
   if (sigBLead.size() > 0 && sigBTrail.size() > 0) {
     for (unsigned i = 0; i < sigBLead.size() && i < sigBTrail.size(); i++) {
```
kkacprzak: similarly for second loop

LargeBarrelAnalysis/EventCategorizerTools.cpp
```diff
@@ -179,15 +180,41 @@ double EventCategorizerTools::calculateTOT(const JPetHit& hit)
                    .getPoints(JPetSigCh::Trailing, JPetRawSignal::ByThrNum);
   auto sigBTrail = hit.getSignalB().getRecoSignal().getRawSignal()
                    .getPoints(JPetSigCh::Trailing, JPetRawSignal::ByThrNum);
-
+
```
alekgajos: Stray spaces:-)

LargeBarrelAnalysis/EventCategorizerTools.h
```diff
@@ -31,14 +31,21 @@ static const double kUndefinedValue = 999.0;
 class EventCategorizerTools
 {
 public:
+  enum TOTCalculationType
+  {
+    kStandard,
+    kAdjustedToThresholds,
+    kAdjustedToThresholdsExtended
```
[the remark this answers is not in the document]
kdulski: I think this is good idea, but third method is estimating it more like a trapeze, so maybe third name would be kThresholdTrapeze?

LargeBarrelAnalysis/EventCategorizerTools.cpp
kdulski: Agree

LargeBarrelAnalysis/EventCategorizerTools.cpp
```diff
@@ -179,15 +180,41 @@ double EventCategorizerTools::calculateTOT(const JPetHit& hit)
...
+      }
+      else 
+        weight = 1.;
+      tot += weight*(sigATrail.at(i).getValue() - sigALead.at(i).getValue());
     }
   }
   if (sigBLead.size() > 0 && sigBTrail.size() > 0) {
     for (unsigned i = 0; i < sigBLead.size() && i < sigBTrail.size(); i++) {
-      tot += (sigBTrail.at(i).getValue() - sigBLead.at(i).getValue());
+      if( i>0 ) {
```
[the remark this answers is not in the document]
kdulski: You`re right, I introduce additional function calculateTOTside in the next commit

PR #154, opened 2020-05-06
Ci actions develop

wkrzemien:
YML Actions configuration file. Also, commented out Image reconstruction since it relies on some file still not presented in the official framework


wkrzemien:
@alekgajos  done

PR #155, opened 2020-05-26
TOT calculations - correction to consider missing thresholds

kdulski:
Following http://sphinx.if.uj.edu.pl/redmine/issues/1349
TOT calculations corrected if some threshold are missing.

### DIFF
--- a/LargeBarrelAnalysis/EventCategorizerTools.cpp
+++ b/LargeBarrelAnalysis/EventCategorizerTools.cpp
@@ -186,20 +186,23 @@ double EventCategorizerTools::calculateTOTside(const std::map<int, double> & thr
   if (!thrToTOT_side.empty()) {
     double firstThr = thrToTOT_side.begin()->first;
     tot += weight*it->second;
-    for (auto it = thrToTOT_side.begin(); it != thrToTOT_side.end(); ++it) {
-      switch(type) {
-        case TOTCalculationType::kSimplified:
-          weight = 1.;
-          break;
-        case TOTCalculationType::kThresholdRectangular:
-          weight = (it->first - std::prev(it, 1)->first)/firstThr;
-          break;
-        case TOTCalculationType::kThresholdTrapeze:
-          weight = (it->first - std::prev(it, 1)->first)/firstThr;
-          tot += weight*(it->second - std::prev(it, 1)->second)/2;
-          break;
+    if( thrToTOT_side.size() > 1 )
+    {
+      for (auto it = std::next(thrToTOT_side.begin(), 1); it != thrToTOT_side.end(); ++it) {
+        switch(type) {
+          case TOTCalculationType::kSimplified:
+            weight = 1.;
+            break;
+          case TOTCalculationType::kThresholdRectangular:
+            weight = (it->first - std::prev(it, 1)->first)/firstThr;
+            break;
+          case TOTCalculationType::kThresholdTrapeze:
+            weight = (it->first - std::prev(it, 1)->first)/firstThr;
+            tot += weight*(it->second - std::prev(it, 1)->second)/2;
+            break;
+        }
+        tot += weight*it->second;
       }
-      tot += weight*it->second;
     }
   }
   return tot;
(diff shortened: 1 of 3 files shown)


LargeBarrelAnalysis/EventCategorizerTools.cpp
```diff
@@ -171,41 +171,34 @@ double EventCategorizerTools::calculateTOT(const JPetHit& hit, TOTCalculationTyp
...
 }
 
-double EventCategorizerTools::calculateTOTside(const std::vector<JPetSigCh> & leadPoints, const std::vector<JPetSigCh> & trailPoints, TOTCalculationType type)
+double EventCategorizerTools::calculateTOTside(const std::map<int, double> & thrToTOT_side, TOTCalculationType type)
 {
   double tot = 0., weight = 1.;
-  if (leadPoints.size() > 0 && trailPoints.size() > 0) {
-    tot += (trailPoints.at(0).getValue() - leadPoints.at(0).getValue());
-    for (unsigned i = 1; i < leadPoints.size() && i < trailPoints.size(); i++) {
+  double firstThr = std::next(thrToTOT_side.begin(), 1)->first;
```
alekgajos, line 186: I think this should be `double firstThr = thrToTOT_side.begin()->first`

LargeBarrelAnalysis/EventCategorizerTools.cpp
```diff
@@ -171,41 +171,34 @@ double EventCategorizerTools::calculateTOT(const JPetHit& hit, TOTCalculationTyp
...
+  if (!thrToTOT_side.empty()) {
+    for (auto it = thrToTOT_side.begin(); it != thrToTOT_side.end(); ++it) {
       switch(type) {
         case TOTCalculationType::kSimplified:
           weight = 1.;
           break;
         case TOTCalculationType::kThresholdRectangular:
-          weight = (leadPoints.at(i).getThreshold() - leadPoints.at(i-1).getThreshold())/leadPoints.at(0).getThreshold();
+          weight = (it->first - std::prev(it, 1)->first)/firstThr;
```
alekgajos, line 194: `it` starts from `thrToTOT_side.begin()` so std::prev(it,1) can reach to the -1st element of the map, which means undefined behaviour.

LargeBarrelAnalysis/EventCategorizerTools.cpp
```diff
@@ -171,41 +171,34 @@ double EventCategorizerTools::calculateTOT(const JPetHit& hit, TOTCalculationTyp
...
           break;
         case TOTCalculationType::kThresholdTrapeze:
-          weight = (leadPoints.at(i).getThreshold() - leadPoints.at(i-1).getThreshold())/leadPoints.at(0).getThreshold();
-          tot += weight*((trailPoints.at(i).getValue() - leadPoints.at(i).getValue()) - 
-                    (trailPoints.at(i-1).getValue() - leadPoints.at(i-1).getValue()))/2;
+          weight = (it->first - std::prev(it, 1)->first)/firstThr;
```
alekgajos, line 197: also here

LargeBarrelAnalysis/EventCategorizerTools.cpp
kdulski: Hmm, but in loop I already increment before the loop (++it), so it should not be a problem here. I just saw that i miss calculation of tot on the first threshold. Should be corrected in the next commit.

alekgajos: Note that ++it in the last part of the for loop does not increment before the 1st iteration. This is always the instruction performed at the end of each iteration, and ++it or it++ does not matter except possibly for performance.

kdulski: Ok, my mistake. I corrected the loop, so it always start from the second element, and added a condition before the loop to check wheter there is at least 2 elements in map. If only one function returns TOT from the first threshold.

alekgajos: Thank you. Does it compile now though? It seems that the test builds are failing.

kdulski: In next commit it should be all right

PR #157, opened 2020-06-01
Added scripts for parallel analysis

Alvarness:
I added python script that runs analysis in parallel. It requires 2 packages that are not instaled by default:
tqdm and termcolor
both of which should be on the server.
First script runs any framework-examples executable and analyzes all files in a specified directory.

Second script purges all root from parmbanks files in specified directory by running purge.C script made by Alek.

In case ./script.py -h is not enough I will prepare readme with detailed description of how to run them

### DIFF
--- a/scripts/parallel_purge.py
+++ b/scripts/parallel_purge.py
@@ -5,19 +5,6 @@ import argparse
 from subprocess import Popen, PIPE
 from multiprocessing.dummy import Pool as PoolThread
 
-try:
-    from termcolor import colored
-except ImportError:
-    print("\033[93m" + "Please instal module termcolor for Python3. \n \
-run command: sudo apt-get install python3-termcolor \n \
-or equivalent on your operating system" + "\033[0m")
-try:
-    import tqdm
-except ImportError:
-    print("\033[93m" + "Please instal module tqdm for Python3. \n \
-run command: sudo apt-get install python3-tqdm \n \
-or equivalent on your operating system" + "\033[0m")
-
 
 def main():
     parser = argparse.ArgumentParser(
@@ -35,20 +22,29 @@ def main():
     progress_bar = args["progress_bar"]
     threads = args["number_of_threads"]
 
+    try:
+        import tqdm
+    except ImportError:
+        print(
+            "\033[93m" + "Module tqdm for Python3 not found. Running without progress bar" + "\033[0m")
+        progress_bar = False
+
     if threads > 20:
-        print(colored(
-            "Try not to use more than 20 threads, let others also run analysis.", "red", attrs=["underline"]))
+        print("\033[31m" + "Try not to use more than 20 threads, let others also run analysis." +
+              "\033[0m", attrs=["underline"])
         exit()
 
     if not path.isdir(input_directory):
-        print(colored(
-            "Specified input drectory des not exist. Please check spelling.", "red"))
+        print(
+            "\033[31m" + "Specified input drectory des not exist. Please check spelling." + "\033[0m")
         exit()
 
     if input_directory[-1] != "/":
         input_directory += "/"
 
     list_of_files = filter(listdir(input_directory), "*.root")
+
+    print("\033[32m" + "All checks passed, purging now." + "\033[0m")
 
     pool = PoolThread(threads)
 
--- a/scripts/parallel_purge.py
+++ b/scripts/parallel_purge.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python3
 from fnmatch import filter
-from os import system, path, listdir
+from os import path, listdir, popen
 import argparse
 
 from multiprocessing.dummy import Pool as PoolThread
@@ -53,7 +53,7 @@ def main():
     pool = PoolThread(threads)
 
     def run_macro_parallel(file):
-        system("root -l -b -q \"purge.C(\\\"{}{}\\\")\"".format(input_directory, file))
+        popen("root -l -b -q \"purge.C(\\\"{}{}\\\")\"".format(input_directory, file))
 
     if progress_bar:
         for _ in tqdm.tqdm(pool.imap(run_macro_parallel, list_of_files), total=len(list_of_files)):
--- a/scripts/parallel_purge.py
+++ b/scripts/parallel_purge.py
@@ -1,9 +1,9 @@
 #!/usr/bin/env python3
 from fnmatch import filter
-import os
-import sys
-import subprocess
+from os import system, path, listdir
 import argparse
+
+from multiprocessing.dummy import Pool as PoolThread
 
 try:
     from termcolor import colored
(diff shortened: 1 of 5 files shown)


scripts/run_parallel_analysis.py
```diff
@@ -0,0 +1,105 @@
...
+
+from multiprocessing.dummy import Pool as PoolThread
+from os import listdir, system, path
+from sys import exit
+from fnmatch import filter
+from termcolor import colored
+import tqdm
+import argparse
+
+
```
wkrzemien, line 11: could you add the main function and move the code there?

wkrzemien:
you could add requirements.txt file with tqdm and termcolor. This is a typical practice in python projects.

scripts/parallel_purge.py
```diff
@@ -0,0 +1,66 @@
...
+    try:
+        import tqdm
+    except ImportError:
+        print(
+            "\033[93m" + "Module tqdm for Python3 not found. Running without progress bar" + "\033[0m")
+        progress_bar = False
+
+    if threads > 20:
+        print("\033[31m" + "Try not to use more than 20 threads, let others also run analysis." +
+              "\033[0m", attrs=["underline"])
```
wkrzemien, line 34: My checker calls syntax error in line 34. Can you please check it

scripts/run_parallel_analysis.py
```diff
@@ -0,0 +1,142 @@
...
+            ", ".join(supported_extensions[1:])) + "\033[0m")
+        exit()
+
+    if output_directory is not None:
+        def run_analysis_parallel(filename):
+            system("./{} -t root -f {} -p conf_trb3.xml -u userParams.json -i {} -l detectorSetupRun{}.json -o {}".format(
+                executable, filename, run_id, run_id_setup, output_directory))
+
+    else:
+        def run_analysis_parallel(i):
```
wkrzemien, line 117: i is never used. If you really want use _

wkrzemien: but which filename do you use in that case?

scripts/run_parallel_analysis.py
```diff
@@ -0,0 +1,142 @@
...
+    supported_extensions = ["*", "hld", "hld.root", "tslot.calib.root", "raw.sig.root",
+                            "phys.sig.root", "hits.root", "unk.evt.root", "cat.evt.root"]
+
+    if extension not in supported_extensions:
+        print("\033[31m" + "Specified file extension is not valid. Please check if it's one of the following: \n{}".format(
+            ", ".join(supported_extensions[1:])) + "\033[0m")
+        exit()
+
+    if output_directory is not None:
+        def run_analysis_parallel(filename):
```
wkrzemien, line 112: please define external (to main) and one function run_parallel_analysis() which takes the needed arguments, output_directory can be optional.

wkrzemien: The function should be called run_parallel_analysis to be consistent with the file name

scripts/run_parallel_analysis.py
wkrzemien: i think it is undefined

scripts/run_parallel_analysis.py
```diff
@@ -0,0 +1,142 @@
...
+
+    print("\033[32m" + "All checks passed, running analysis now." + "\033[0m")
+
+    pool = PoolThread(threads)
+
+    if progress_bar:
+        for _ in tqdm.tqdm(pool.imap(run_analysis_parallel, list_of_files), total=len(list_of_files)):
+            pass
+    else:
+        results = pool.map(run_analysis_parallel, list_of_files)
```
wkrzemien, line 135: Results is not used anywhere

scripts/run_parallel_analysis.py
```diff
@@ -0,0 +1,142 @@
...
+    input_directories = [directory + "/" if directory[-1] !=
+                         "/" else directory for directory in input_directories]
+
+    if output_directory is not None:
+        if not path.isdir(output_directory):
+            print(
+                "\033[31m" + "Specified output drectory des not exist. Please check spelling or create a directory." + "\033[0m")
+            exit()
+
+        if output_directory[-1] is not "/":
```
wkrzemien, line 79: You should use `!= ` and no `is not`, since you want to compare the content , and not the object identity here

scripts/run_parallel_analysis.py
```diff
@@ -0,0 +1,142 @@
...
+    extension = args["extension"]
+
+    try:
+        import tqdm
+    except ImportError:
+        print(
+            "\033[93m" + "Module tqdm for Python3 not found. Running without progress bar" + "\033[0m")
+        progress_bar = False
+
+    run_id_setup = run_id
```
wkrzemien, line 50: I would extract a small function. Something like:
```
def get_run_id_setup(run_id):
  run6_mapping = {"61": "6A",
                    "62": "6B",
                    "63": "6C",
                    "64": "6D"}
  if not run_id:
     some error
   if run_id[0] == "6":
     return run_mapping[run_id]
return run_id
```

scripts/run_parallel_analysis.py
```diff
@@ -0,0 +1,142 @@
...
+        if not path.isdir(output_directory):
+            print(
+                "\033[31m" + "Specified output drectory des not exist. Please check spelling or create a directory." + "\033[0m")
+            exit()
+
+        if output_directory[-1] is not "/":
+            output_directory += "/"
+
+    if file_type != "root":
+        allowed_types = ["root", "mcGeant", "hld", "zip", "scope"]
```
wkrzemien, line 83: maybe those, could be defined as global variable `ALLOWED_TYPES` ?

scripts/parallel_purge.py
```diff
@@ -0,0 +1,66 @@
+#!/usr/bin/env python3
+from fnmatch import filter
+from os import path, listdir, popen
+import argparse
+from subprocess import Popen, PIPE
```
wkrzemien, line 5: Popen is not used I think

wkrzemien: ok it is

scripts/parallel_purge.py
```diff
@@ -0,0 +1,66 @@
...
+
+    def run_macro_parallel(file):
+        Popen("root -l -b -q \"purge.C(\\\"{}{}\\\")\"".format(input_directory,
+                                                               file), shell=True, stdout=PIPE).wait()
+
+    if progress_bar:
+        for _ in tqdm.tqdm(pool.imap(run_macro_parallel, list_of_files), total=len(list_of_files)):
+            pass
+    else:
+        result = pool.map(run_macro_parallel, list_of_files)
```
wkrzemien, line 59: result is not used

scripts/parallel_purge.py
```diff
@@ -0,0 +1,66 @@
+#!/usr/bin/env python3
+from fnmatch import filter
+from os import path, listdir, popen
```
wkrzemien, line 3: this popen is not used

scripts/parallel_purge.py
```diff
@@ -0,0 +1,66 @@
...
+        import tqdm
+    except ImportError:
+        print(
+            "\033[93m" + "Module tqdm for Python3 not found. Running without progress bar" + "\033[0m")
+        progress_bar = False
+
+    if threads > 20:
+        print("\033[31m" + "Try not to use more than 20 threads, let others also run analysis." +
+              "\033[0m", attrs=["underline"])
+        exit()
```
wkrzemien, line 35: shouln't it be sys.exit()

scripts/run_parallel_analysis.py
Alvarness: I strongly disagree:
1) Only place this function will be used is from main function so there is no necessity to define it globally
2) By defining it in main it has access to all command line variables without necessity to pass them
3) By defining 2 different versions we are moving conditional statement outside of function definition. Function is defined once, either with specified output or not - otherwise we would check condition for very file instead of once

wkrzemien: Sorry, but those are totally wrong arguments in view of programming practices and general guideline how you should organize code.
Ad1. You could also claim that since most of the code is used only in main so there is no sense to extract any function at all.  
Ad 2. That is completely wrong. Passing arguments makes it explicit and clear what is the input and otput to the functions. This is exactly the problem why you made a mistake with i/filename
Ad3. I dont understand this argument

Alvarness: Ad 1) There is nothing wrong with defining function inside function
Ad 2) I remembered why I did it way I did - passing more than one argument to pool.map is bit painful. I pass all the parameters now, you can check if it's more clear in this version
Ad 3) Right now (after change) we are checking whether output it specified with every call of parallel_analysis function so a couple thousands of times. I wanted to move this if out of function to check it once and execute appropriate version

Regardless, I moved parallel_analysis function outside of main and it's working now

wkrzemien: Ok, thanks. It is better I think.
I just added some modifications and I opened as the PR to your code. Please check it, It should solve  point 3.

wkrzemien: https://github.com/Alvarness/j-pet-framework-examples/pull/1/files

PR #158, opened 2020-06-01
Extend the MCGeantAnalysis example

alekgajos:
Extensions of the example include:
- example of controlling MC smearing with user parameters
- creating an analysis module that works on both MC and data
- basic accessing of generated MC information
- documentation updates related to all of the above.

### DIFF
--- a/MCGeantAnalysis/EventAnalyzer.cpp
+++ b/MCGeantAnalysis/EventAnalyzer.cpp
@@ -29,11 +29,11 @@ bool EventAnalyzer::init() {
   INFO("Event analysis started.");
 
   getStatistics().createHistogramWithAxes(
-      new TH1D("z_res", "Resolution along Z", 100, -15., 15.),
+      new TH1D("z_res", "Resolution along Z", 301, -15.05, 15.05),
       "Z_{REC}-Z_{MC} [cm]");
 
   getStatistics().createHistogramWithAxes(
-      new TH1D("Edep_res", "Resolution of deposited energy", 100, -200., 200.),
+      new TH1D("Edep_res", "Resolution of deposited energy", 201, -201., 201.),
       "E_{REC}-E_{MC} [keV]");
 
   // Input events type
--- a/MCGeantAnalysis/EventAnalyzer.cpp
+++ b/MCGeantAnalysis/EventAnalyzer.cpp
@@ -28,12 +28,13 @@ EventAnalyzer::~EventAnalyzer() {}
 bool EventAnalyzer::init() {
   INFO("Event analysis started.");
 
-  getStatistics().createHistogram(new TH1F(
-      "z_res", "Resolution along Z;Z_{REC}-Z_{MC} [cm]", 100, -15., 15.));
+  getStatistics().createHistogramWithAxes(
+      new TH1D("z_res", "Resolution along Z", 100, -15., 15.),
+      "Z_{REC}-Z_{MC} [cm]");
 
-  getStatistics().createHistogram(new TH1F(
-      "Edep_res", "Resolution of deposited energy;E_{REC}-E_{MC} [keV]", 100,
-      -200., 200.));
+  getStatistics().createHistogramWithAxes(
+      new TH1D("Edep_res", "Resolution of deposited energy", 100, -200., 200.),
+      "E_{REC}-E_{MC} [keV]");
 
   // Input events type
   fOutputEvents = new JPetTimeWindow("JPetEvent");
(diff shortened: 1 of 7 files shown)


MCGeantAnalysis/EventAnalyzer.cpp
```diff
@@ -0,0 +1,100 @@
...
+ * limitations under the License.
+ *
+ *  @file EventAnalyzer.cpp
+ */
+
+#include "EventAnalyzer.h"
+#include "../LargeBarrelAnalysis/EventCategorizerTools.h"
+#include <JPetMCHit/JPetMCHit.h>
+#include <JPetOptionsTools/JPetOptionsTools.h>
+#include <iostream>
```
wkrzemien, line 20: do you need iostream?

MCGeantAnalysis/run.sh
```diff
@@ -5,4 +5,4 @@ if [ -z ${3+x} ]; then
     exit 
 fi
 
-./MCGeantAnalysis.x -t mcGeant -f $1 -u userParams.json -l $2 -i $3
+./MCGeantAnalysis.x -t mcGeant -f $1 -u userParams.json -l $2 -i $3 -u userParams.json
```
kdulski, line 8: Do i see it right? Two times -u userParams.json?

MCGeantAnalysis/EventAnalyzer.cpp
```diff
@@ -0,0 +1,100 @@
...
+    // for each reconstructed hit, we access the corresponding
+    // "true MC" hit
+    const JPetMCHit &mc_hit =
+        tw->getMCHit<JPetMCHit>(reconstructed_hit.getMCindex());
+
+    getStatistics().getHisto1D("z_res")->Fill(reconstructed_hit.getPos().Z() -
+                                              mc_hit.getPos().Z());
+    getStatistics()
+        .getHisto1D("Edep_res")
+        ->Fill(reconstructed_hit.getEnergy() - mc_hit.getEnergy());
```
kdulski, line 98: Maybe it will be useful to also use new methods for filling histograms :) ?
and creating for line 32-37?

MCGeantAnalysis/EventAnalyzer.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+EventAnalyzer::EventAnalyzer(const char *name) : JPetUserTask(name) {}
+
+EventAnalyzer::~EventAnalyzer() {}
+
+bool EventAnalyzer::init() {
+  INFO("Event analysis started.");
+
+  getStatistics().createHistogram(new TH1F(
+      "z_res", "Resolution along Z;Z_{REC}-Z_{MC} [cm]", 100, -15., 15.));
```
kdulski, line 33: Hmm z can be negative and therefore it can produce some artifact in 0 if the bins are defined like here.
Maybe define histogram bins like: 
300, -14.95, 15.05
or 
301, -15, 15

MCGeantAnalysis/EventAnalyzer.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+  getStatistics().createHistogram(new TH1F(
+      "Edep_res", "Resolution of deposited energy;E_{REC}-E_{MC} [keV]", 100,
+      -200., 200.));
```
kdulski, line 37: Similar comment as above
Bins like
400, -199.5, 200.5
or
401, -200, 200

kdulski:
Some minor comments. It is very good idea to have an example how to treat simulated data by the Framework. It will resolve many problems.

MCGeantAnalysis/EventAnalyzer.cpp
alekgajos: Right, of course not. I have removed it and reorganized the includes.

MCGeantAnalysis/run.sh
alekgajos: Yes, you are right. I don't know how I managed to do it:-)
Deleted.

MCGeantAnalysis/EventAnalyzer.cpp
alekgajos: Right, I have transformed the example to use the new interface.

MCGeantAnalysis/EventAnalyzer.cpp
alekgajos: Fixed.

kkacprzak:
Just a comment: if one would like to be pedantic about words, it would be better to use "MC Data" and "Measurement data"

alekgajos:
> Just a comment: if one would like to be pedantic about words, it would be better to use "MC Data" and "Measurement data"

I see your point but I'm a little afraid that using "MC data" will cause confusion as we commonly use "data" for what is not MC. Let us discuss it at the meeting tomorrow.

wkrzemien:
@alekgajos @kkacprzak  I agree here with Alek "MC data" is misleading.

PR #168, opened 2020-08-06
TOT calculation to HitFinder / adding flag for method of TOT calculation

kdulski:
Moving methods for TOT calculations to HitFinder

Adding flag for selection of the method of the TOT calculations


wkrzemien:
@kdulski  You probably need to modify cmakes as well. So far the tests are not passing.

wkrzemien:
@kdulski because of the undefined references

CosmicAnalysis/userParams.json
```diff
@@ -0,0 +1,24 @@
+{
```
wkrzemien, line 1: Im not 100% sure, but I think we don't add userParams.json to the repo.

CosmicAnalysis/PARAMETERS.md
```diff
@@ -34,6 +34,9 @@ time window for matching Signals on the same scintillator and different sides. D
 - `HitFinder_RefDetScinID_int`  
 `ID` of Reference Detector Scintillator, needed for creating reference hits
 
+- `HitFinder_TOTCalculationType_std::string`  
+Type of the calculations of the TOT - it can be standard sum (option "standard"), a extended sum taking into account thresholds differences and calculated as rectangulars (option "rectangular"), additional extension that add also differences between the TOTs on different thresholds and calculates sum as sum of the trapezes (option "trapeze")
```
wkrzemien, line 38: Is there a default value?

LargeBarrelAnalysis/HitFinderTools.cpp
```diff
@@ -267,31 +269,59 @@ void HitFinderTools::checkTheta(const double& theta)
...
 
 /**
-* Calculation of the total ToT of the hit - Time over Threshold:
-* the sum of the ToTs on all of the thresholds (1-4) and on the both sides (A,B)
+* Calculation of the total TOT of the hit - Time over Threshold:
+* the weighted sum of the TOTs on all of the thresholds (1-4) and on the both sides (A,B) 
 */
-double HitFinderTools::calculateToT(const JPetHit& hit)
+
+void HitFinderTools::setTOTCalculationType(std::string type)
```
wkrzemien, line 276: `const std::string& type`

LargeBarrelAnalysis/HitFinderTools.cpp
```diff
@@ -267,31 +269,59 @@ void HitFinderTools::checkTheta(const double& theta)
...
+{
+    if (type == "rectangular") {
```
wkrzemien, line 278: You don't check the  "standard" keyword which should be used I think? Also, probably some warning/error should be given in case of other type is used

LargeBarrelAnalysis/HitFinderTools.h
```diff
@@ -59,7 +65,12 @@ class HitFinderTools
   static JPetHit createDummyRefDetHit(const JPetPhysSignal& signal);
   static int getProperChannel(const JPetPhysSignal& signal);
   static void checkTheta(const double& theta);
-  static double calculateToT(const JPetHit& hit);
+  static void setTOTCalculationType(std::string type);
+  static double calculateTOT(const JPetHit& hit);
+  static double calculateTOTside(const std::map<int, double> & thrToTOT_side);
+private:
+    
+  static TOTCalculationType fTOTCalculationType;
```
wkrzemien, line 73: I don't like this member here, it introduces some state of the class, which is not needed. . I would remove it,
And do something like that:
1.Define a function that changes a string to enum:
`static TOTCalculationType getToTCalculationType(const std::string& type); // here the body of setTOT`
but with return at the end.
2.Add explicit argument of `TOTCalculationType`  to calculateTOT e.g:
`static double calculateTOT(const JPetHit& hit, TOTCalculationType type = kSimplified);`
3.When calling calculateTOT, one can use  getToTCalculationType to get the second argument

CosmicAnalysis/userParams.json
kdulski: fixed. Where it should be added?

LargeBarrelAnalysis/HitFinderTools.h
kdulski: corrected in the new PR

LargeBarrelAnalysis/HitFinderTools.cpp
kdulski: Ok, check added

LargeBarrelAnalysis/HitFinderTools.cpp
kdulski: applied

CosmicAnalysis/PARAMETERS.md
kdulski: added default value : 'standard'

CosmicAnalysis/userParams.json
alekgajos: The json parameter files are downloaded from the server. I will update them if you let me know what I should change - just add 
```
"HitFinder_TOTCalculationType_std":"standard"
```
?

alekgajos: I have updated all of the userParams.json files at the server.

alekgajos (changes requested):
Good work! Just two minor comments (in the code) and one general doubt to discuss:
I don't know which is the right approach myself but just started wondering if examples such as PhysicsAnalysis, CosmicsAnalysis and Imaging shouldn't use one of the more advanced TOT calculation methods by default?
What do you think?

kdulski:
Well, if we use more sophisticated way of calculating TOT in, it will require more control from the user, for example to properly set the thresholds. I would stand by default one, to leave other options for experienced users.

alekgajos:
You are right about the need for properly provided thresholds in such case. I agree that the standard calculation should be default but what concerns me is that if we also use the default parameter of `calculateTOT`, users may not even see from these examples that there is a possibility to do it in a more sophisticated way. Perhaps it would be better to copy the same user option from the general EventCategorizer to these examples as well so that it's visible that there are other options?

alekgajos:
You are right about the need for properly provided thresholds in such case. I agree that the standard calculation should be default but what concerns me is that if we also use the default parameter of `calculateTOT`, users may not even see from these examples that there is a possibility to do it in a more sophisticated way. Perhaps it would be better to copy the same user option from the general EventCategorizer to these examples as well so that it's visible that there are other options?

alekgajos:
You are right about the need for properly provided thresholds in such case. I agree that the standard calculation should be default but what concerns me is that if we also use the default parameter of `calculateTOT`, users may not even see from these examples that there is a possibility to do it in a more sophisticated way. Perhaps it would be better to copy the same user option from the general EventCategorizer to these examples as well so that it's visible that there are other options?

alekgajos:
You are right about the need for properly provided thresholds in such case. I agree that the standard calculation should be default but what concerns me is that if we also use the default parameter of `calculateTOT`, users may not even see from these examples that there is a possibility to do it in a more sophisticated way. Perhaps it would be better to copy the same user option from the general EventCategorizer to these examples as well so that it's visible that there are other options?

alekgajos:
You are right about the need for properly provided thresholds in such case. I agree that the standard calculation should be default but what concerns me is that if we also use the default parameter of `calculateTOT`, users may not even see from these examples that there is a possibility to do it in a more sophisticated way. Perhaps it would be better to copy the same user option from the general EventCategorizer to these examples as well so that it's visible that there are other options?

alekgajos:
You are right about the need for properly provided thresholds in such case. I agree that the standard calculation should be default but what concerns me is that if we also use the default parameter of `calculateTOT`, users may not even see from these examples that there is a possibility to do it in a more sophisticated way. Perhaps it would be better to copy the same user option from the general EventCategorizer to these examples as well so that it's visible that there are other options?

wkrzemien (approved):
Thanks. looks good for me

kdulski:
Ok, I think I understand. I will add in Cosmic/Imaging/Physic appropriate flag and TOT calculation type in the next commit.
In addition I updated initialization of the histograms in Cosmic/Imaging/Physic, so they are the same as in LargeBarrelAnalysis

alekgajos (approved):

PR #177, opened 2021-05-16
Add initial version of interfile reconstruction

kamilrakoczy:
This PR adds ability to save image reconstructed using naive (MLP) method in interfile format.

Signed-off-by: Kamil Rakoczy <kamilrakoczy1@gmail.com>


InterfileReconstruction/InterfileMLPReconstruction.cpp
```diff
@@ -0,0 +1,133 @@
...
+    for (unsigned int i = 0; i < numberOfEventsInTimeWindow; i++)
+    {
+      auto event = dynamic_cast<const JPetEvent &>(timeWindow->operator[](static_cast<int>(i)));
+      auto numberOfHits = event.getHits().size();
+      if (numberOfHits <= 1)
+        continue;
+      else
+      {
+        auto hits = event.getHits();
+        for (unsigned int i = 0; i < hits.size() - 1; i++)
```
wkrzemien, line 40: I think this code is not 100% correct.
Lets assume you have 4 hits, then what you want to do is to combine all the possibilities right?
E.g. (1,2), (1,3), (1,4) etc. but you will combine only (1,2) and (2,3) 
I think you should use double for e.g.
`for (unsigned int i = 0; i < hits.size() ; i++) {
for (unsigned int j = i+1; j< hits.size() ; j++) {
TVector3 annihilationPoint = EventCategorizerTools::calculateAnnihilationPoint(hits[i], hits[j]);
}
}`

InterfileReconstruction/InterfileMLPReconstruction.cpp
```diff
@@ -0,0 +1,133 @@
...
+  auto x_bins = pos_histo->GetNbinsX();
+  auto y_bins = pos_histo->GetNbinsY();
+  auto z_bins = pos_histo->GetNbinsZ();
+
+  //Intefile format is defined in a way, that we saves 2D arrays
+  //from top left corner to bottom right
+  //in row major order, more information in article:
+  //https://www.researchgate.net/publication/21707960_A_file_format_for_the_exchange_of_nuclear_medicine_image_data_A_specification_of_interfile_version_33
+
+  for (auto i = 0; i < z_bins; i++) {
```
wkrzemien, line 72: for the sake of testability I would extract this code to a stand-alone function eg. `bool saveTofile(TH3D* hist, const char* outFileName);`

wkrzemien: Then it can go to tools

InterfileReconstruction/InterfileMLPReconstruction.h
```diff
@@ -0,0 +1,36 @@
+#ifndef INTERFILEMLPRECONSTRUCTION_H
+#define INTERFILEMLPRECONSTRUCTION_H
+
+#include "JPetUserTask/JPetUserTask.h"
+#include <memory>
+
```
wkrzemien, line 6: Please, add two lines of description what this class does.

wkrzemien (commented):
Very nice and clean!
Few suggestions added.

InterfileReconstruction/InterfileMLPReconstruction.cpp
wkrzemien: or better `saveToInterflie (TH3D* image, const std::string& outFileName);`

InterfileReconstruction/InterfileMLPReconstruction.cpp
kamilrakoczy: I added second loop.

InterfileReconstruction/InterfileMLPReconstruction.cpp
kamilrakoczy: Done.

InterfileReconstruction/InterfileMLPReconstruction.h
kamilrakoczy: Done.

wkrzemien (approved):
Nice work!

PR #180, opened 2021-06-23
Correcting trapeze method - proper adding triangle

kdulski:
There was a mistake in calculation of the TOT by trapeze method - wrong sign of the addition of the triangle component.

Fixed by adding fabs function, as it will be more general, when we do not know which TOTs will be smaller.


wkrzemien (approved):
For me it looks fine :)

PR #184, opened 2022-04-08
Tot synchronization

nenprio:
I have added the possibility of doing the TOT synchronization using a given calibTOT.json file with the corresponding constants to the HitFinder and HitFinderTools to the LargeBarrelAnalysis code. 
To solve the problem of keeping the value of the synchronized TOT's for future modules/analysis, I am storing it with setEnergy(). Later in the analysis one can always use getEnergy() to have sync. TOT.
PARAMETERS.md has been updated with the two new fields: a boolean to decide whether to sync. and the path to the calibration file.
Also the 2D histogram required for the sync. scripts (to be updated) is added to HitFinder (tot_per_scin).


LargeBarrelAnalysis/HitFinderTools.h
```diff
@@ -68,6 +70,8 @@ class HitFinderTools
   static TOTCalculationType getTOTCalculationType(const std::string& type);
   static double calculateTOT(const JPetHit& hit, TOTCalculationType type = kSimplified);
   static double calculateTOTside(const std::map<int, double> & thrToTOT_side, TOTCalculationType type);
+  static void saveTOTsync(std::vector<JPetHit>& hits,  std::string TOTCalculationType, boost::property_tree::ptree fSyncTree);
```
alekgajos, line 73: The last parameter could probably be a const reference to `boost::property_teree::ptree`.
Moreover, could you change the parameter name to e.g. syncTree? Starting with 'f' is a convention we use to indicate class member variables, so it'll be confusing in case of a function parameter.

LargeBarrelAnalysis/HitFinderTools.h
```diff
@@ -68,6 +70,8 @@ class HitFinderTools
...
+  static double syncTOT(const JPetHit& hit, double TOT, boost::property_tree::ptree fSyncTree);
```
alekgajos, line 74: Please see the two comments for `saveTOTsync` above as they also apply here.

LargeBarrelAnalysis/PARAMETERS.md
```diff
@@ -63,6 +63,12 @@ time window for matching Signals on the same scintillator and different sides. D
 - `HitFinder_TOTCalculationType_std::string`  
 Type of the calculations of the TOT - it can be standard sum (option "standard"), a extended sum taking into account thresholds differences and calculated as rectangulars (option "rectangular"), additional extension that add also differences between the TOTs on different thresholds and calculates sum as sum of the trapezes (option "trapeze"). Default value: 'standard'
 
+- `HitFinder_SyncToT_bool`
+Boolean to decide to do the TOT syncrhonization. True/False. If true syncFile.json is required.
```
alekgajos, line 67: Please change "syncFile.json" to e.g. "a calibration file indicated with the `TOTConstantsFile_std::string`" option (see below).

alekgajos (changes requested):
Please try to reformat the modified files according to our formatting rules using the `clang-format` tool. You can call is as:
```
clang-format -i --style=file file_to_be_formatted.cpp
```
This way it will use the settings from the `.clang-format` file located in the framework root directory to format the file consistently with the rest of code.

alekgajos (approved):

PR #185, opened 2022-04-28
New module for downscaling events with certain numbers of hits

alekgajos:
This PR adds the "Downscaler" module to LargeBarrelAnalysis.

"Downscaler" is a simple filtering module which randomly accepts events with certain number of hits at certain rates provided by the user using the `Downscaler_DownscalingRates_std::vector<double>` parameter (see the changes in `PARAMETERS.md` for its interpretation).

The module is needed for the upcoming Run 11 preselection.

By the way, this PR adds the previously-missing `.clang-format` file (same as in j-pet-framework) to the repository root dir.

### DIFF
--- a/LargeBarrelAnalysis/Downscaler.cpp
+++ b/LargeBarrelAnalysis/Downscaler.cpp
@@ -20,72 +20,80 @@
 using namespace jpet_options_tools;
 using namespace std;
 
-Downscaler::Downscaler(const char *name) : JPetUserTask(name) {
+Downscaler::Downscaler(const char* name) : JPetUserTask(name)
+{
   std::random_device rd;
   fRandomGenerator = std::mt19937(rd());
 }
 
 Downscaler::~Downscaler() {}
 
-bool Downscaler::init() {
+bool Downscaler::init()
+{
   INFO("Event downscaling started.");
 
   // Parameter for back to back categorization
-  if (isOptionSet(fParams.getOptions(), fDownscalingRatesKey)) {
-    fDownscalingRates =
-        getOptionAsVectorOfDoubles(fParams.getOptions(), fDownscalingRatesKey);
-  } else {
+  if (isOptionSet(fParams.getOptions(), fDownscalingRatesKey))
+  {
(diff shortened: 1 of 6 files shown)


LargeBarrelAnalysis/Downscaler.cpp
```diff
@@ -0,0 +1,98 @@
...
+Downscaler::~Downscaler() {}
+
+bool Downscaler::init()
+{
+  INFO("Event downscaling started.");
+
+  // Parameter for back to back categorization
+  if (isOptionSet(fParams.getOptions(), fDownscalingRatesKey))
+  {
+    fDownscalingRates = getOptionAsVectorOfDoubles(fParams.getOptions(), fDownscalingRatesKey);
```
wkrzemien, line 38:
```suggestion
    fDownscalingRates = getOptionAsVectorOfDoubles(fParams.getOptions(), fDownscalingRatesKey);
    std::transform(fDownscalingRates.begin(), fDownscalingRates.end(), fDownscalingRates.begin(), [](double el)->double { return el/100.;});
```
Since later you always use values divided by 100, maybe it would be better to divide it already here?

LargeBarrelAnalysis/Downscaler.cpp
```diff
@@ -0,0 +1,98 @@
...
+    vector<JPetEvent> events;
+    for (uint i = 0; i < timeWindow->getNumberOfEvents(); i++)
+    {
+      const auto& event = dynamic_cast<const JPetEvent&>(timeWindow->operator[](i));
+
+      int multiplicity = event.getHits().size();
+      double rate = 1.0;
+      if (multiplicity <= fDownscalingRates.size())
+      {
+        rate = fDownscalingRates[multiplicity - 1] / 100.;
```
wkrzemien, line 67: Then you don't need /100 anymore

wkrzemien (changes requested):
Nice work. One small suggestion added, but if you don't want, you can skip it.

wkrzemien (approved):

PR #190, opened 2023-06-07
TOT converter only when requested

kkacprzak:
This PR aims to solve the issue of creating TOT converter only when necessary, as reported in this [redmine](http://sphinx.if.uj.edu.pl/redmine/issues/1919) issue,

The ToT converter is loaded with parameters only if requested and the options are provided, otherwise, the tot factory is null and does not return generated converted, so no error messages should be logged.

The tests for Hit Finder Tools were updated and are working.

### DIFF
--- a/LargeBarrelAnalysis/ToTEnergyConverterFactory.cpp
+++ b/LargeBarrelAnalysis/ToTEnergyConverterFactory.cpp
@@ -1,5 +1,5 @@
 /**
- *  @copyright Copyright 2020 The J-PET Framework Authors. All rights reserved.
+ *  @copyright Copyright 2023 The J-PET Framework Authors. All rights reserved.
  *  Licensed under the Apache License, Version 2.0 (the "License");
  *  you may not use this file except in compliance with the License.
  *  You may find a copy of the License in the LICENCE file.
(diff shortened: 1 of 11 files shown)


LargeBarrelAnalysis/ToTEnergyConverterFactory.cpp
```diff
@@ -76,8 +76,28 @@ void ToTEnergyConverterFactory::loadConverterOptions(const ToTEnergyConverterFac
   {
     fToT2EnergyAll = {toT2EnergyFormula, {toT2EnergyParameters, {}}};
   }
+
+  /// Converter factory by default is null, only after proper initialisation
+  /// the flag is changed to true and can be used in the examples
+  fIsNullObject = false;
 }
 
+void ToTEnergyConverterFactory::setEnergyConverterOptions(const FuncParamsAndLimits formula)
```
wkrzemien, line 85: Why do we need those setters? Is it not possible to provide the formula while creating the factory?

LargeBarrelAnalysis/ToTEnergyConverterFactory.cpp
```diff
@@ -76,8 +76,28 @@ void ToTEnergyConverterFactory::loadConverterOptions(const ToTEnergyConverterFac
...
+{
+  fEnergy2ToTAll = formula;
+  fIsNullObject = false;
+}
+
+void ToTEnergyConverterFactory::setToTConverterOptions(const FuncParamsAndLimits formula)
+{
+  fToT2EnergyAll = formula;
+  fIsNullObject = false;
```
wkrzemien, line 94: I think there is a logical problem how you define that the factory is ready.  You provide the formula  to the directions (tot2Energy or Energy2Tot) , maybe instead of boolean it would be good to have some kind of enum that marks that the factory is not ready, it is ready for e2Tot, it is ready for tot2e or both?

wkrzemien (changes requested):
Nice work! I am not sure what is the logic behind the error handling. e.g. if fIsNullObject set to false, does it mean that either of two converters can be created? It would be good to know which one is ready.
I would suggest to add  to unit tests such cases where: (1) the formula is wrong (e.g. syntactically) what should happen in that case ? (2)You set the formula with one of the setter (e..g ToT2Energy and you call getEnergyConverter() and the opposite. What would happen there?

PR #195, opened 2025-10-01
Introducing trigger signal to MTAB example

kkacprzak:
Changes for MTAB example, including histograms for the time of signals/hits with or without trigger.


kkacprzak (commented):
Changes introduce the flag for trigger/no-tirgger time windows, add histograms for hit and signal time distribution and hit time modified by trigger time.

